PR #51, opened 2022-04-24
workaround bugs in clang and gcc to optimize compiled efficiency

### LINKED ISSUE #50: 🫠Fix compiled output efficiency regressions

lukevalenty:
resolves #50


lukevalenty:
this fixes the compiled output efficiency...but the fix reverts a compile-time performance enhancement. will need to come up with a new solution to speed up the compile-time performance of the cib::service<T> pointers.

PR #58, opened 2022-05-01
implement indexed_tuple with different get method

lukevalenty:


include/cib/detail/indexed_tuple.hpp
```diff
@@ -0,0 +1,153 @@
...
+
+    template<typename IndexMetafuncT0, typename... Ts>
+    [[nodiscard]] constexpr auto make_indexed_tuple(index_metafunc_<IndexMetafuncT0>, Ts const & ... values) {
+        return make_indexed_tuple(std::make_integer_sequence<int, sizeof...(values)>{}, index_metafunc_<IndexMetafuncT0>{}, values...);
+    }
+
+
+    template<typename Callable, typename... Elements>
+    constexpr auto apply(Callable operation, indexed_tuple<Elements...> const & t) {
+        return operation(t.Elements::value...);
```
lukevalenty, line 101: clang doesn't think there is a pack to expand here, but gcc has no problem with this.

lukevalenty: worked around this by moving the apply implementation into the indexed_tuple class

lukevalenty:
need to update set.hpp to implement constexpr set operations so it is compatible with c++17.

PR #59, opened 2022-05-03
fix and simplify find.hpp

lukevalenty:


lukevalenty:
Pull request #58 solves this problem in a much better way :)

PR #61, opened 2022-05-07
:racehorse: :art: [cib/detail] Replace `type_pack_element` recursive …

krzysztof-jusiak:
…version

Problem:
- When `__type_pack_element` builtin is not implemented `type_pack_element`
  is using recursion which is slow to compile.

Solution:
- Speed up compilation time by replacing non-recursive version of type_pack_element when
  `__type_pack_element` is not available by creating inherited list of type
  and id pairs and leverage inheritance to get the Nth type pack
  element.

### DIFF
--- a/include/cib/detail/type_pack_element.hpp
+++ b/include/cib/detail/type_pack_element.hpp
@@ -1,4 +1,5 @@
 #include <type_traits>
+#include <utility>
 
 
 #ifndef COMPILE_TIME_INIT_BUILD_TYPE_PACK_ELEMENT_HPP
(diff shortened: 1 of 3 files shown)


include/cib/detail/type_pack_element.hpp
```diff
@@ -1,23 +1,29 @@
...
 
 #ifndef COMPILE_TIME_INIT_BUILD_TYPE_PACK_ELEMENT_HPP
 #define COMPILE_TIME_INIT_BUILD_TYPE_PACK_ELEMENT_HPP
 
 
 namespace cib::detail {
-    #if defined(__clang__)
+    #if not defined(__clang__)
         template<auto Index, typename... Tn>
         using type_pack_element = __type_pack_element<Index, Tn...>;
```
lukevalenty, line 12: I think the conditional check for clang is inverted. All of the unit tests are failing compilation with gcc because of this.

lukevalenty (changes requested):
thank you for your pull request @krzysztof-jusiak! 😊

i approved your account to execute the github actions. once the unit tests are passing i will be happy to merge this.

include/cib/detail/type_pack_element.hpp
krzysztof-jusiak: Good catch @lukevalenty, thanks, should be fixed now

lukevalenty (approved):

test/detail/type_pack_element.cpp
```diff
@@ -0,0 +1,16 @@
+#include <cib/detail/type_pack_element.hpp>
```
lukevalenty, line 1: there's a unit test build for clang and gcc that uses the single-header <cib/cib.hpp> header. all of the unit tests should include that instead of the specific files.  this will take care of the last two unit test failures.

PR #93, opened 2022-07-27
change Catch2 test dependency to conan package

lukevalenty:


lukevalenty:
hmmm, conan isn't sitting quite right with me at the moment for header-only libraries.

PR #95, opened 2022-08-18
add flow, container, and string_constant libraries

lukevalenty:
some shiny new features being included as part of the cib library. this pull request is the first of several building up cib into a framework.


include/container/Array.hpp
```diff
@@ -0,0 +1,152 @@
...
+                "Initializer list size {} is bigger than allocated array size {}",
+                src.size(), int_<Size>);
+
+        } else {
+            auto dst_index = 0;
+            for (auto v : src) {
+                storage[dst_index] = v;
+                dst_index++;
+            }
+        }
```
miscco, line 50: This will lead to an additional copy. 

The forst copy happens in the ranged for loop, the second copy in the assignment.

For generic nontrivial data, this should capture by `const&`

include/container/Array.hpp
```diff
@@ -0,0 +1,152 @@
...
+        : storage{}
+    {
+        if (src.size() > Size) {
+            FATAL(
+                "Source size {} is bigger than allocated array size {}",
+                src.size(), int_<Size>);
+
+        } else {
+            auto dst_index = 0;
+            for (auto v : src) {
```
miscco, line 64: Ditto additional copy.

include/container/ConstexprMap.hpp
```diff
@@ -0,0 +1,257 @@
...
+ *
+ * @tparam KeyType
+ * @tparam ValueType
+ */
+template<
+    typename KeyType,
+    typename ValueType>
+struct MapEntry {
+    KeyType key;
+    ValueType value;
```
miscco, line 21: The elements are always value-initialized.

Any reason to not do this as member initializer?

lukevalenty: Don't have a good reason. 😅 However, this container is intended to only be used in a constexpr context. I think it should still be fixed though.

PR #108, opened 2022-08-30
add message library and handler service

lukevalenty:
Adding the message library and handler service. This was presented by @mjcaisse-intel at cpp now 2022.

https://www.youtube.com/watch?v=LdPgNFR7OL8


test/msg/field.cpp
```diff
@@ -0,0 +1,210 @@
+#include <catch2/catch_test_macros.hpp>
+
+#include <msg/field.hpp>
+
+namespace msg {
+    using TestField1 = field<
+        decltype("TestField1"_sc),
+        0, 15, 0,
```
bryceschober, line 8: Suggestion: I'm a fancy-C++ novice, but I think I've seen a technique used elsewhere (Maybe the Physical units for matrices talk from CppNow?) that would allow me to use a named-parameter-style idiom here that would be way more expressive, along the lines of:
```c++
using TestField1 = field<
        decltype("TestField1"_sc),
        WordIdx<0>, Msb<15>, Lsb<0>,
        std::uint32_t>;
```

PR #123, opened 2022-09-12
Update CMakelists, remove git submodule command and updated README

siauthun:
1. Update CMakelists -Added new lib/CMakeList.txt
2. Remove git submodule command in root's CMakeList
3. Updated README on user have to manually perform git submodule when using cib as a standalone project


.gitmodules
```diff
@@ -1,6 +0,0 @@
-[submodule "lib/Catch2"]
```
lukevalenty, line 1: we need to keep the .gitmodules file, no need to remove it

README.md
```diff
@@ -116,9 +116,14 @@ ls build/include/cib/ | grep cib.hpp
...
 
-**NOTE:** *cib* uses git submodules to include its testing dependencies. The 
-CMake configuration *should* fetch the submodules for you, but only if the
-repository was cloned as a git repo and not downloaded as an archive.
+**NOTE:** When *cib* is cloned as standalone project, users are required to manually
+fetch ```Catch2``` and ```fmt``` submodules into ```lib/Catch2``` and ```lib/fmt```
+as per command below. These two libraries are needed due to *cib* testing dependencies.
+
+```shell
+git submodule add https://github.com/catchorg/Catch2.git lib/Catch2
```
lukevalenty, line 124: we don't want them to add the git submodules themselves, the .gitmodules file contains this information already. instead they can run the following command:

```shell
git submodule update --recursive
```
```

lukevalenty (changes requested):

lukevalenty:
the unit tests will also need to pass before merging :)

siauthun:
Screw up the commits. Unable to revert deleted lib/Catch2 and lib/fmt. Will create another pull request

PR #126, opened 2022-09-13
Sequencer addition

kkwisele:
Adding new Sequencer service similar to Flow but it allows forward and backward execution of the steps.


CMakeLists.txt
```diff
@@ -13,7 +13,7 @@ project(
     DESCRIPTION "A header-only C++ library for composing modular firmware at compile-time."
     HOMEPAGE_URL "https://github.com/intel/compile-time-init-build")
 
-add_library(cib INTERFACE)
+add_library(cib INTERFACE include/seq/impl.hpp)
```
lukevalenty, line 16: do we need to add the "include/seq/impl.hpp? we should remove this if it's not necessary.

lukevalenty (approved):

bdeane-intel:
Superseded by #206

PR #136, opened 2022-09-25
addition of interrupt manager

lukevalenty:


lukevalenty:
this is a work in progress. need to do the following tasks still:
1) use cib::tuple instead of hana
2) clean up the mock register interface
3) add readme and examples
4) ensure both catch2 and gtest tests run 
5) spend some time to clean up the file/directory structure

lukevalenty:
added initial interrupt manager

Issue #27, opened 2022-04-13
🕵🏽‍♀️❓ Investigate why gcc-11 fails to compile the unit tests in C++20 mode


lukevalenty:
gcc-11 fails to build the unit tests when it is run under C++20 mode. We need to understand whether this is a GCC bug or bug in cib. If it's a GCC bug, we need to workaround the issue and file a bug on GCC.

PR #148, opened 2022-09-29
Fix create_type_names for GCC 12.2

elbeno:
Immediate (consteval) functions can't be called from constexpr functions. There is no reason for create_type_names not to be an immediate function, I think.

Note: re https://github.com/intel/compile-time-init-build/issues/27, I think this is related. The reason tests don't compile properly under GCC 11.2 (and now 12.2) is likely that consteval functions are being called from constexpr functions (or lambdas, which is the same thing). See also https://gcc.gnu.org/bugzilla/show_bug.cgi?id=95968 - lambdas are implicitly constexpr (if they can be) but not implicitly immediate. I don't think this is a GCC bug.

With this change, all tests compile with GCC 12.2 except nexus.cpp and flow.cpp, which both run into the constexpr/consteval issue related to tuple handling.

The Array test compiles but fails at https://github.com/intel/compile-time-init-build/blob/main/test/container/Array.cpp#L46. Not sure why this changes with GCC 12.2, but zero-size C-style arrays are not standard in C++...

### DIFF
--- a/include/cib/detail/set_details.hpp
+++ b/include/cib/detail/set_details.hpp
@@ -76,7 +76,7 @@ namespace cib::detail {
     };
 
     template<typename Tag>
-    constexpr static std::string_view name() {
+    CIB_CONSTEVAL static std::string_view name() {
         #if defined(__clang__)
             constexpr std::string_view function_name = __PRETTY_FUNCTION__;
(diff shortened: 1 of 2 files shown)


lukevalenty (approved):
thanks @elbeno! :)

PR #163, opened 2022-10-10
Use CRTP for type-based tuple indexing

elbeno:
Remove multiple `tuple_element` specializations for indexing with types; replace it with a CRTP mixin that allows indexing with types in a parameter pack.


lukevalenty (approved):
looks great!

PR #191, opened 2022-10-26
moving over to cpm vs submodules

lukevalenty:


cmake/libraries.cmake
```diff
@@ -0,0 +1,29 @@
+function(target_link_library_includes target scope)
```
lukevalenty, line 1: praise: i like this function!

PR #198, opened 2022-11-04
:rotating_light: Turn on warnings

bdeane-intel:
Warnings are applied to all unit test targets. Warnings are moved to a separate interface library where they can be easily seen and augmented. Warnings now apply to every toolchain; for some reason before they were turned on only for clang-12.

Of course, targets now compile cleanly without warnings. This involved a few code changes.

Turning on warnings caught at least 2 bugs:
- failure to actually call an IILE, meaning the constexpr bool it was assigned to was always true.
- an infinite loop caused by incorrect loop conditions on an unsigned loop variable.


lukevalenty (approved):
i _really_ like these changes

PR #199, opened 2022-11-04
:bug: Fix constexpr evaluation for GCC 11

bdeane-intel:
GCC 11 correctly points out that the following cannot be compiled:

```cpp
struct S {
  int i{42};
  consteval auto func() { return i; }
};

constexpr auto call_func_on(auto obj) {
  return obj.func();
}

constexpr S s;
constexpr auto x = call_func_on(s);
```

Even though `s` is `constexpr`, inside `call_func_on` it's a parameter, and there's no such thing as constexpr function parameters in C++. So attempting to call a `consteval` function fails. When func is `constexpr`, everything is OK.

GCC 11 also correctly spots that indexing into a C-style array of size 0 is UB, so that test is changed to merely declaring an array of size 0. TBD: `Array<T, 0>` should simply not support many operations. See [what the standard says about `std::array<T, 0>`](http://eel.is/c++draft/array.zero).

### DIFF
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -49,12 +49,11 @@ jobs:
             install: sudo apt update && sudo apt-get install -y clang-11
             cc: " /usr/lib/llvm-11/bin/clang"
             cxx: "/usr/lib/llvm-11/bin/clang++"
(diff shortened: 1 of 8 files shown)


mjcaisse-intel (approved):
Nice!

PR #200, opened 2022-11-04
:art: Inject ConcurrencyPolicy into dynamic_controller

bdeane-intel:
It's better to pass a policy as a template argument than rely on a global dependency.

### DIFF
--- a/test/interrupt/manager.cpp
+++ b/test/interrupt/manager.cpp
@@ -3,6 +3,8 @@
 #include <gmock/gmock.h>
 #include <gtest/gtest.h>
 
+#include <conc/concurrency.hpp>
+
 using ::testing::_;
 using ::testing::InSequence;
 using ::testing::Return;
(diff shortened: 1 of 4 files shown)


test/interrupt/manager.cpp
```diff
@@ -275,7 +277,7 @@ struct NoIsrBuilder {
                              rsp_handler_irq, policies<>>>,
                  irq<38, 0, timer_irq, policies<>>>;
 
-        manager<Config> builder;
+        manager<Config, test::ConcurrencyPolicy> builder;
```
mjcaisse-intel, line 280: Maybe a `using` alias for `manager` with the `test::ConcurrencyPolicy`?

mjcaisse-intel (approved):

lukevalenty (approved):

test/interrupt/manager.cpp
bdeane-intel: Good call.

PR #201, opened 2022-11-07
:construction_worker: :rotating_light: Turn on clang-tidy for quality…

bdeane-intel:
… target

Turning on more clang-tidy checks is a work in progress. This is a good start. Among the changes here:
- applying NSDMIs, often resulting in default constructors becoming unnecessary or able to use `= default`
- move construction/assignment should be `noexcept`
- core guideline C.128 (virtual/override/final: specify only one)
- single-argument constructors should be explicit

Note that the mipi catalog encoder has external reliance on definitions of `CIB_LOG_CRITICAL_SECTION_RAII_TYPE` and `CIB_LOG_DESTINATIONS`: that is temporarily worked around here to allow a clean quality build.


mjcaisse-intel (approved):
This is great!

PR #207, opened 2022-11-10
:art: Use #pragma once consistently

bdeane-intel:
Some headers were using `#pragma once`, some were using include guards. This commit changes them all to use `#pragma once`. In addition, some redundant includes were removed from a few places.


mjcaisse-intel (approved):
Like the consistency. Thanks for limiting the header files also.

bdeane-intel:
There are probably more header cleanups that can be done; this was some low-hanging fruit. Including hana in a fwd.hpp was a bit egregious I thought :)

PR #209, opened 2022-11-10
:white_check_mark: Add test for tuple for_each

bdeane-intel:
`for_each` now supports:
- invocation on a const tuple or a mutable tuple -- it can mutate the tuple members (like std::for_each)
- returning the callable that was passed in (also like std::for_each)
- invocation as a free function to run a function over the expansion of several tuples

### DIFF
--- a/include/cib/tuple.hpp
+++ b/include/cib/tuple.hpp
@@ -167,9 +167,14 @@ template <typename... TupleElementTs> struct tuple_impl : TupleElementTs... {
      *      The operation to perform. Must be a callable that accepts a single
      * parameter.
      */
-    template <typename Callable>
-    CIB_CONSTEXPR void for_each(Callable operation) const {
-        (operation(TupleElementTs::value), ...);
+    template <typename TOp>
+    CIB_CONSTEXPR auto for_each(TOp &&op) const & -> TOp {
+        (op(TupleElementTs::value), ...);
+        return op;
+    }
+    template <typename TOp> CIB_CONSTEXPR auto for_each(TOp &&op) & -> TOp {
+        (op(TupleElementTs::value), ...);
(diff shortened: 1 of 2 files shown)


include/cib/tuple.hpp
```diff
@@ -167,9 +167,14 @@ template <typename... TupleElementTs> struct tuple_impl : TupleElementTs... {
      *      The operation to perform. Must be a callable that accepts a single
      * parameter.
      */
-    template <typename Callable>
-    CIB_CONSTEXPR void for_each(Callable operation) const {
-        (operation(TupleElementTs::value), ...);
+    template <typename TOp>
+    CIB_CONSTEXPR auto for_each(TOp &&op) const & -> TOp {
```
mjcaisse-intel, line 171: TIL `std::for_each` returns the UnaryFunction

include/cib/tuple.hpp
```diff
@@ -167,9 +167,14 @@ template <typename... TupleElementTs> struct tuple_impl : TupleElementTs... {
...
+        (op(TupleElementTs::value), ...);
+        return std::forward<TOp>(op);
+    }
+    template <typename TOp> CIB_CONSTEXPR auto for_each(TOp &&op) & -> TOp {
+        (op(TupleElementTs::value), ...);
+        return std::forward<TOp>(op);
```
mjcaisse-intel, line 177: Does this defeat any implicit move or RVO that might happen?

bdeane-intel: RVO isn't happening because it's a parameter. But I think it's a good call on the implicit move.

mjcaisse-intel (approved):

lukevalenty (approved):

PR #210, opened 2022-11-10
:fire: Remove unused containers

bdeane-intel:
DoubleLinkedList, LinkedList and Queue are entirely unused except for their own test code. We don't need to maintain them and can always bring them back in future if needed.


mjcaisse-intel (approved):
Woot! My favorite type of reviews.

PR #211, opened 2022-11-14
:construction_worker: Use Ninja for CI, rework build matrix

bdeane-intel:
- Remove support for clang 7 & 8 -- they have problems with some constexpr constructs that will be hard to support in future.
- Use Ninja for CI builds -- this shaves about 2 minutes off the quality build.
- Rework build matrix to support all desired combinations: compilers, compiler versions, C++ standards, build type (for now only Debug).
- `single_header` builds are removed for now because they don't actually make a difference? We don't have tests that include `cib/cib.hpp` as a unified header.

Note: ubuntu-latest is currently 20.04, NOT 22.04 yet. Hence GCC 12+ and Clang 15+ are not available without moving forward either by updating to ubuntu-22.04 or when github updates ubuntu-latest to 22.04.

### DIFF
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -8,7 +8,7 @@ on:
     branches: [ main ]
 
 env:
-  BUILD_TYPE: Debug
+  CMAKE_GENERATOR: Ninja
 
 jobs:
   build_and_test:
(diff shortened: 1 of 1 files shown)


mjcaisse-intel (approved):
👍

PR #213, opened 2022-11-14
:fire: Remove Array

bdeane-intel:
Array is not needed: we use std::array in many places, and `std::array` provides more functionality (like the ability to declare a zero-size array).

The only member of `std::array` that can throw exceptions is `at`, and we don't use that.


lukevalenty (approved):
👍

mjcaisse-intel (approved):

PR #214, opened 2022-11-14
:art: Remove CIB_CONSTEXPR

bdeane-intel:
We target C++17 and above; `CIB_CONSTEXPR` is never anything other than `constexpr`. `constexpr` is already used a lot and we might as well use the standard keyword everywhere.

This commit also cleans up the include usage for `cib/details/compiler.hpp`.


mjcaisse-intel (approved):
🧹  clean clean clean

lukevalenty (approved):
looks good to me! :)

not sure why I added CIB_CONSTEXPR now that I think about it 🤔

PR #215, opened 2022-11-15
:bug: Use libfmt in header-only mode

bdeane-intel:
Linking against `fmt::fmt` causes projects that depend on cib to build the non-header parts of libfmt, which cib doesn't use. Better to link against `fmt::fmt-header-only`.


lukevalenty (approved):
🧹

rubyliuintelcom (approved):

PR #222, opened 2022-11-23
Fix Logger uses dynamic memory (std::string)

Thurnheer:
fmt::format returns a string which uses dynamic memory so every trace and log produced small dynamic memory allocations switch to use fmt::format_to with fmt::memory_buffer thsize can be set with changing fmt::internal::INLINE_BUFFER_SIZE
closes#221

### DIFF
--- a/include/log/fmt/logger.hpp
+++ b/include/log/fmt/logger.hpp
@@ -39,10 +39,11 @@ template <> struct fmt::formatter<log_level> {
 };
 
 namespace {
-inline const auto loggingStartTime = std::chrono::high_resolution_clock::now();
+inline const auto logging_start_time =
+    std::chrono::high_resolution_clock::now();
 
 template <typename StringType>
-inline auto trimSourceFilename(StringType src) -> char const * {
+inline auto trim_source_filename(StringType src) -> char const * {
     return std::strstr(src, "src");
 }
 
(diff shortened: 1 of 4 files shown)


include/log/fmt/logger.hpp
```diff
@@ -56,25 +57,28 @@ template <typename T> struct FormatHelper {
...
 
 template <typename FilenameStringType, typename LineNumberType,
           typename MsgType>
-void log(FilenameStringType filename, LineNumberType lineNumber,
+void log(FilenameStringType filename, LineNumberType line_number,
          log_level level, MsgType msg) {
-    auto formattedMsg = msg.args.apply(
-        [&](auto... args) { return fmt::format(msg.str.value, args...); });
+    auto formatted_msg = msg.args.apply([&](auto... args) {
+        auto out = fmt::memory_buffer();
```
lukevalenty, line 76: question: where is this memory allocated? on the stack? There is no parameter for a buffer size, how big of a buffer is it creating? i couldn't find the documentation for fmt::memory_buffer.

include/log/fmt/logger.hpp
```diff
@@ -39,10 +39,11 @@ template <> struct fmt::formatter<log_level> {
 };
 
 namespace {
-inline const auto loggingStartTime = std::chrono::high_resolution_clock::now();
+inline const auto logging_start_time =
+    std::chrono::high_resolution_clock::now();
 
 template <typename StringType>
-inline auto trimSourceFilename(StringType src) -> char const * {
+inline auto trim_source_filename(StringType src) -> char const * {
```
lukevalenty, line 46: thanks for  fixing the naming here! :)

lukevalenty (changes requested):
Thanks for your pull request! :) I have some questions on how fmt::memory_buffer works that I need to understand before we can merge these changes.

include/log/fmt/logger.hpp
lukevalenty: ahh, your original comment answers my question. i should read more carefully!

> ...set with changing fmt::internal::INLINE_BUFFER_SIZE

lukevalenty (approved):

lukevalenty:
> Is there any issuue with thisPR? Would you prefer a different solution?

It's all good, I was missing some information on `fmt::memory_buffer`, but I found it in your original PR message. :)

PR #227, opened 2022-11-30
🍜 Enable more clang tidy checks

bdeane-intel:
Remove usage of std::tuple, C-style arrays, and raw indexing.


mjcaisse-intel (approved):
⚔️  slaying the c-array

lukevalenty (approved):
😋 🍜

include/flow/impl.hpp
```diff
@@ -61,15 +63,15 @@ template <typename Name, int NumSteps> class impl : public interface {
      *
      * @see flow::builder
      */
-    constexpr impl(milestone_base newMilestones[], build_status newBuildStatus)
+    constexpr impl(milestone_base *newMilestones, build_status newBuildStatus)
         : functionPtrs(), buildStatus(newBuildStatus) {
         if constexpr (loggingEnabled) {
-            for (int i = 0; i < NumSteps; i++) {
+            for (auto i = std::size_t{}; i < NumSteps; i++) {
```
bryceschober, line 69: Question from the peanut gallery: When looking through your work, this caught my eye. Is there a particular reason to do this as opposed to `std::size_t i = 0`?

bryceschober: I'm in a season of considering coding standards in a modern C++ context, so that's why I'm probing - Thanks for the principled explanations!

Issue #231, opened 2022-12-09
Having issue on multiple lambda callback and WithIn format on enum class


YinThong:
Created test case with similar condition in a #230

PR #232, opened 2022-12-09
:bug: Fix message handler with multiple callbacks

bdeane-intel:
The way that tuple filter works currently requires the elements of the tuple to be default constructible. This was not the case for `any_t`.

This is a partial fix for #231 (which is two issues in one).


lukevalenty (approved):
thanks @bdeane-intel!

PR #238, opened 2022-12-14
:art: DRY out match::any and match::all

bdeane-intel:
`any_t` and `all_t` do the same thing, just with different monoids. The match test defined its own `always`, but this is available from the header.


mjcaisse-intel (approved):
Looks great!

PR #240, opened 2022-12-19
:art: Allow injection of logger configuration

bdeane-intel:
Logging in general has 3 separately configurable options:
 - Where to send the logs (the destinations)
 - What type to use for a critical section (if needed)
 - How to handle `CIB_FATAL` logs or equivalently `CIB_ASSERT`

Injecting these options makes logging testable and allows overriding behaviour without the preprocessor. The default behaviour is now the null logger. On a per-program basis this can be overridden by specializing the `logging::config` variable template. Suitable configurations are given for fmt logging and for mipi logging.


lukevalenty (approved):
🤩 Much nicer way of picking a logger output without brittle include order dependencies!

PR #262, opened 2023-02-14
:recycle: Refactor cmake infrastructure

bdeane-intel:
- don't download test frameworks if they are not used
- provide dummy clang-tidy targets when not using clang toolchain


mjcaisse-intel (approved):
Looks good!

PR #265, opened 2023-02-28
:construction_worker: Use C++20

bdeane-intel:
- Use C++20 in CI, trim supported compilers to GCC 12, Clang 14, Clang 15


lukevalenty (approved):
🥳 🍾

PR #282, opened 2023-03-21
:sparkles: Log CPM dependency tree

bdeane-intel:
It's useful to have a log file of the CPM dependency tree, to see why a given dependency was/wasn't fulfilled (because it was previously downloaded at an earlier version by a different dependency). This change creates one in `${CMAKE_BINARY_DIR}/cpm_dependencies.txt`.


mjcaisse-intel (approved):
Nice!

PR #284, opened 2023-03-22
:recycle: Improve circular queue

bdeane-intel:
- made the API closer to std::queue, more useful
- added a policy for safe over/underflow


nikhilshahani (approved):
We may need to consider concurrency use-cases in the future.

PR #290, opened 2023-04-21
Update README.md

nanobug:
I got confused when I read the example .md and did not understand how the output shows 'Don't panic' without referencing the component, then saw the code .cpp has it.


bdeane-intel (approved):
Good catch.

PR #293, opened 2023-04-21
C++20 cleanup to use `constexpr` STL algorithms

AshleyRoll:
Update to use C++20 constexpr std algorithms

- update code tagged to be rewritten after updating to C++20.
- convert Vector to use C++20 algorithms
- convert message to use C++20 algorithms
- convert sc/format to C++20 algorithms

The change to `flow::generic_builder::hasNoIncomingEdges()` resulted in
a build failure in GCC, but success in Clang. This was tracked down to
the `begin()` and `end()` iterators some how resulting in a different
type in GCC. This then caused GCC's implementation to fail when it
attempted to subtract the iterators.

This was fixed by always using the `std::{c}begin()` method to generate
both the begin and end iterators. Other occurrences of the generation of
end iterators in a similar way were identified and corrected in
`Vector`.

### DIFF
--- a/include/sc/detail/string_meta.hpp
+++ b/include/sc/detail/string_meta.hpp
@@ -2,6 +2,7 @@
 
 #include <sc/fwd.hpp>
 
+#include <algorithm>
 #include <array>
 #include <string_view>
 #include <utility>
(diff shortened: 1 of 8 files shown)


include/sc/detail/string_meta.hpp
```diff
@@ -24,28 +25,22 @@ struct Replace {
...
 
         auto const first = thisStr.begin() + pos;
         auto const last = first + count;
 
         // copy begin() until first
-        for (auto src = thisStr.begin(); src != first; src++, dst++) {
-            *dst = *src;
-        }
+        std::copy(thisStr.begin(), first, dst);
+        std::advance(dst, pos);
```
bdeane-intel, line 36: **suggestion**: `copy` returns the output iterator, so `dst = std:copy(...)` ?

include/sc/detail/string_meta.hpp
```diff
@@ -24,28 +25,22 @@ struct Replace {
...
 
         // copy str.begin() until str.end()
-        for (auto src = str.begin(); src != str.end(); src++, dst++) {
-            *dst = *src;
-        }
+        std::copy(str.begin(), str.end(), dst);
+        std::advance(dst, str.length());
```
bdeane-intel, line 40: And of course here too.

include/sc/detail/string_meta.hpp
```diff
@@ -24,28 +25,22 @@ struct Replace {
...
 
         // copy last until end()
```
bdeane-intel, line 42: **thought**: Given the code is now one-liners that are the same as the 3 comments in this function, I think we can remove the comments.

bdeane-intel (commented):
Nice - I'd been meaning to get around to this!

include/sc/detail/string_meta.hpp
AshleyRoll: doh.. I did that elsewhere and forgot to fix it here. Fix incoming.

include/sc/detail/string_meta.hpp
AshleyRoll: Yep, agree. wasn't sure about your commenting policy so I left them there. Will remove.

AshleyRoll:
For some reason the auto-merge wasn't happening. I thought it might be because the branch was out of date with `main` so I tried rebasing it. Looks like that removed the approvals. GH Actions are running very very slowly today it seems for this project.

PR #297, opened 2023-04-25
:fire: Abstract the flow graph builder and use it for seq also

bdeane-intel:
Flow and seq libraries use practically identical code for building graphs, so separate this into a flow::graph_builder and have them both use the same thing.


lukevalenty (approved):
Looks great 👨‍🍳

PR #308, opened 2023-05-18
:sparkles: Add ability to destructure cib::tuple

### LINKED ISSUE #307: cib::tuple should be destructurable
Need to add tests (& functionality) so that e.g. the following works:

```cpp
auto const t = cib::tuple{1, 3.14f};
auto const [i, f] = t;
CHECK(i == 1);
CHECK(f == 3.14f);
```

bdeane-intel:
Destructuring is in a separate header to avoid otherwise unwarranted inclusion of `<tuple>`.

Closes #307


lukevalenty (approved):
Cool! 😎

PR #309, opened 2023-05-19
Support noexcept message callbacks

svdschoot-wefabricate:


lukevalenty (approved):
Thanks for the PR! 😎 We have an open issue to move to a boost library for function traits that _should_ take care of this. However, I'm not sure when we'll get around to it, so we can approve this PR in the meantime.

PR #312, opened 2023-05-31
:recycle: Remove `msg::message_data`; replace with `cib::vector`

bdeane-intel:


lukevalenty (approved):
Yup, I like it. Much nicer to use cib::vector and get rid of message_data. Thanks Ben!

PR #320, opened 2023-07-12
:bug: :boom: Fix message_base size

### LINKED ISSUE #317: `message_base` constructor and copy constructor should be returning sizeof...(ArgFields) instead of `NumDWordsT`?
# Issue 1
It will be better if we could retrieve runtime size of message instead of getting message size from compile time `NumDWordsT` value.
With that, `NumDWordsT` is not in used and can be removed.

# Issue 2
Also in `include/msg/message.hpp` line 119: static_assert should be ensuring size of argFields does not exceed MaxNumDWords.
```
static_assert (sizeof...(ArgFields) <= MaxNumDWords); 
```

bdeane-intel:
`message_base` size was not set correctly when constructing from a range or a variadic pack of integral types.

Remove message_base NumDWords template parameter: there was no need for it.

Closes #317

### DIFF
--- a/include/msg/message.hpp
+++ b/include/msg/message.hpp
@@ -55,11 +55,11 @@ template <typename MsgType, typename AdditionalMatcher>
     return is_valid_msg_t<MsgType, AdditionalMatcher>{};
 }
 
-template <typename NameType, std::uint32_t MaxNumDWords,
-          std::uint32_t NumDWordsT, typename... FieldsT>
+template <typename NameType, std::uint32_t MaxNumDWords, typename... FieldsT>
 struct message_base : public message_data<MaxNumDWords> {
     constexpr static NameType name{};
-    constexpr static auto NumDWords = NumDWordsT;
+    constexpr static auto max_num_dwords = MaxNumDWords;
+    static_assert((... and (FieldsT::MaxDWordExtent < MaxNumDWords)));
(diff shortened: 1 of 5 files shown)


include/msg/message.hpp
```diff
@@ -55,11 +55,11 @@ template <typename MsgType, typename AdditionalMatcher>
...
 }
 
-template <typename NameType, std::uint32_t MaxNumDWords,
-          std::uint32_t NumDWordsT, typename... FieldsT>
+template <typename NameType, std::uint32_t MaxNumDWords, typename... FieldsT>
 struct message_base : public message_data<MaxNumDWords> {
     constexpr static NameType name{};
-    constexpr static auto NumDWords = NumDWordsT;
+    constexpr static auto max_num_dwords = MaxNumDWords;
+    static_assert((... and (FieldsT::DWordIndex < MaxNumDWords)));
```
lukevalenty, line 62: issue: 64-bit fields will have an index less than MaxNumDWords, but the upper bits if the field will be equal to or greater than MaxNumDWords.

lukevalenty (changes requested):
need to fix the check for fields to make sure both the least significant bit and most significant bit of all fields fit in the message data. current check in this PR only ensures the LSB fits within the message data.

mjcaisse-intel (approved):

lukevalenty (approved):

PR #337, opened 2023-08-12
:art: Use `std::function` CTAD for function traits

elbeno:


lukevalenty (approved):
I loooove to see code deleted!

PR #338, opened 2023-08-14
:art: Change format to use fmtlib compile-time formatting

### LINKED ISSUE #257: :bug: formatting a string with compile-time arguments should always return a string_constant and not lazy_string_format

elbeno:
Closes #257
Addresses #105


lukevalenty:
I'm going to need to take a close look at this on Monday. 🤩

lukevalenty (approved):
Great! Would be nice to be able to treat string_constant and lazy_string_format the same way.

PR #344, opened 2023-08-20
:art: Simplify `builder_meta` machinery

elbeno:
- simplify exposing `builder_t` and `interface_t`
- remove vestigial `traits` namespace


include/cib/builder_meta.hpp
```diff
@@ -19,8 +19,14 @@ namespace cib {
  *
  * @example cib::callback_meta
  */
-template <typename BuilderType, typename InterfaceType> struct builder_meta {
-    auto builder() -> BuilderType;
-    auto interface() -> InterfaceType;
+template <typename Builder, typename Interface> struct builder_meta {
+    using builder_t = Builder;
```
lukevalenty, line 23: I used static methods to ensure subclasses would inherit them. When I tried using type aliases, there were cases in which I couldn't access them from a subclass.

lukevalenty (approved):

include/cib/builder_meta.hpp
elbeno: Hmm there are no tests that show this? I can't think immediately why that would work and not this?

PR #345, opened 2023-08-20
:art: Refactor nexus demux algorithm

elbeno:
Demux is a tuple-based version of sort followed chunk_by.


lukevalenty:
Have you checked compilation times for the compilation performance test?

elbeno:
Yes, the big nexus benchmark is about 0.5s more on my laptop. But then some months ago it was about 6s more.

elbeno:
Looking back, it looks like #323 actually impacted the benchmark times quite a bit. Maybe we can take a look at how to bring that in a bit.

elbeno:
#346 should restore compilation perf.

lukevalenty (approved):
*chef's kiss*

PR #347, opened 2023-08-23
:sparkles: Add `contains_type` for `indexed_tuple`

bdeane-intel:


lukevalenty (approved):
Glad to see this get in early!

PR #349, opened 2023-08-24
:bug: Fixed indexed_builder and handler_builder tests

AshleyRoll:
The two tests utilised exactly the name message, field and other types resulting in an ODR violations causing the code to incorrectly call the wrong callback (other file).

Fixed by placing each instance in an anonymous namespace to prevent them leaving the translation unit.


AshleyRoll:
Hi Luke, I wanted to experiment with a different `bitset` but found the test was broken so I fixed it first :)

elbeno:
Oh hi Ashley, actually Luke and I worked this yesterday and we have an updated working branch already too...

AshleyRoll:
Haha, fair enough 😃. Was blocking me so I quickly fixed it. Just close the PR if you don't need the fix.

bdeane-intel:
Thanks for the confirmation - I found the ODR issue last night and fixed the same way! The existing PR should be passing now I think.

bdeane-intel:
OK, closing this one since the original PR is now merged, but thanks, Ashley!

PR #353, opened 2023-08-30
:art: Simplify message catalog types

bdeane-intel:
The types used to encode messages for overload in the catalog don't need to be instantiable. This simplifies what `strings.cpp` needs to include, and what the generation script needs to handle.


lukevalenty (approved):
I like it...providing a clean interface to encode and decode the messages in the catalog.

Issue #370, opened 2023-09-19
💡 `indexed_handler`: support checking for non-indexed fields


lukevalenty:
`indexed_handler` does not support checking for non-indexed fields. 

Fixing this should be relatively straightforward: check the full callback `matcher` before executing the callbacks `callable`.

PR #377, opened 2023-09-19
:bug: Fix issue #370

elbeno:
If a message matches on indexed fields, but doesn't match the full matcher in the callback, the callback shouldn't be called.

### DIFF
--- a/include/msg/handler_interface.hpp
+++ b/include/msg/handler_interface.hpp
@@ -5,7 +5,7 @@ template <typename MsgBaseT, typename... ExtraCallbackArgsT>
 struct handler_interface {
     virtual auto is_match(MsgBaseT const &msg) const -> bool = 0;
 
-    virtual void handle(MsgBaseT const &msg,
-                        ExtraCallbackArgsT... extra_args) const = 0;
+    virtual auto handle(MsgBaseT const &msg,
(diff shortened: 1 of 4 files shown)


test/msg/indexed_handler.cpp
```diff
@@ -41,14 +41,15 @@ template <auto N> using bitset = stdx::bitset<N, std::uint32_t>;
...
 
-TEST_CASE("create empty handler", "[indexed_handler]") {
-    [[maybe_unused]] constexpr auto h = msg::indexed_handler{
-        msg::callback_args<test_msg>,
-        msg::indices{msg::index{
-            opcode_field{}, lookup::make(CX_VALUE(
-                                lookup::input<std::uint32_t, bitset<32>>{}))}},
-        std::array<void (*)(test_msg const &), 0>{}};
-}
+// TEST_CASE("create empty handler", "[indexed_handler]") {
```
lukevalenty: question: why are these tests commented out?

lukevalenty (changes requested):

lukevalenty (approved):

test/msg/indexed_handler.cpp
elbeno: accidental! reverted the commenting :)

PR #379, opened 2023-09-20
:speaker: Add match logging for indexed builder

elbeno:


elbeno:
For now I did not output the extended info when all matches fail; that would probably require a pair of function pointers for the callback (one for match/invoke, one for log_mismatch).

lukevalenty (approved):

PR #380, opened 2023-09-21
:microscope: run length encoded indexed message handling

AshleyRoll:
I have implemented a `rle_indexed_service` and `rle_indexed_builder` to provide a RLE blob version of the existing ones. 

This provides a compressed message field index lookup mechanism to reduce overall storage requirements, especially on large indices.

### DIFF
--- a/include/msg/detail/indexed_builder_common.hpp
+++ b/include/msg/detail/indexed_builder_common.hpp
@@ -1,0 +1,125 @@
+#pragma once
+
+#include <lookup/entry.hpp>
+#include <lookup/input.hpp>
+#include <lookup/lookup.hpp>
+#include <match/ops.hpp>
+#include <msg/field_matchers.hpp>
+
+#include <stdx/bitset.hpp>
+#include <stdx/compiler.hpp>
+#include <stdx/cx_map.hpp>
+#include <stdx/tuple.hpp>
+#include <stdx/tuple_algorithms.hpp>
+
+#include <array>
+#include <cstddef>
+#include <iterator>
+#include <type_traits>
+
+namespace msg {
+template <typename T> using get_field_type = typename T::field_type;
+
+namespace detail {
+template <typename BuilderValue>
+CONSTEVAL auto for_each_callback(auto fn) -> void {
+    constexpr auto t = BuilderValue::value.callbacks;
+    [&]<std::size_t... Is>(std::index_sequence<Is...>) -> void {
+        (fn(t[stdx::index<Is>], Is), ...);
+    }(std::make_index_sequence<t.size()>{});
+}
+
+} // namespace detail
+
+template <typename FieldType, std::size_t EntryCapacity,
+          std::size_t CallbackCapacity>
+struct temp_index {
+    using field_type = FieldType;
+
+    using value_t = stdx::bitset<CallbackCapacity, std::uint32_t>;
+    stdx::cx_map<uint32_t, value_t, EntryCapacity> entries{};
+    value_t default_value{};
+
+    constexpr auto operator[](auto key) -> auto & {
+        if (!entries.contains(key)) {
+            entries.put(key, value_t{});
+        }
+        return entries.get(key);
+    }
+
+    constexpr auto collect_defaults(std::size_t max) {
+        for (auto i = std::size_t{}; i < max; ++i) {
+            if (std::none_of(
+                    std::cbegin(entries), std::cend(entries),
+                    [&](auto const &entry) { return entry.value[i]; })) {
+                default_value.set(i);
+            }
+        }
+    }
+};
+
+template <template <typename, typename, typename, typename...> typename ParentT,
+          typename IndexSpec, typename CallbacksT, typename BaseMsgT,
+          typename... ExtraCallbackArgsT>
+struct indexed_builder_base {
+    CallbacksT callbacks;
+
+    template <typename... Ts> [[nodiscard]] constexpr auto add(Ts... ts) {
+        auto new_callbacks =
+            stdx::tuple_cat(callbacks, stdx::make_tuple(ts...));
+        using new_callbacks_t = decltype(new_callbacks);
+        return ParentT<IndexSpec, new_callbacks_t, BaseMsgT,
+                       ExtraCallbackArgsT...>{new_callbacks};
+    }
+
+    using callback_func_t = void (*)(BaseMsgT const &,
+                                     ExtraCallbackArgsT... args);
+
+    template <typename BuilderValue, std::size_t I>
+    constexpr static auto invoke_callback(BaseMsgT const &msg,
+                                          ExtraCallbackArgsT... args) {
+        // FIXME: incomplete message callback invocation...
+        //        1) bit_cast message argument
+        constexpr auto &orig_cb = BuilderValue::value.callbacks[stdx::index<I>];
+        constexpr auto cb = IndexSpec{}.apply([&]<typename... Indices>(
+                                                  Indices...) {
+            return remove_match_terms<typename Indices::field_type...>(orig_cb);
+        });
+        if (cb.matcher(msg)) {
+            CIB_INFO("Incoming message matched [{}], because [{}], executing "
+                     "callback",
+                     cb.name, orig_cb.matcher.describe());
+            cb.callable(msg, args...);
+        }
+    }
+
+    template <typename BuilderValue, std::size_t... Is>
+    static CONSTEVAL auto create_callback_array(std::index_sequence<Is...>)
+        -> std::array<callback_func_t, BuilderValue::value.callbacks.size()> {
+        return {invoke_callback<BuilderValue, Is>...};
+    }
+
+    template <typename BuilderValue>
+    static CONSTEVAL auto create_temp_indices() {
+        IndexSpec indices{};
+        std::size_t count{};
+        stdx::for_each(
+            [&](auto callback) {
+                build_index(
+                    callback.matcher,
+                    [&]<typename Field, typename V>(std::size_t idx,
+                                                    V expected_value) {
+                        if constexpr (stdx::contains_type<IndexSpec, Field>) {
+                            stdx::get<Field>(indices)[expected_value].set(idx);
+                        }
+                    },
+                    count);
+            },
+            BuilderValue::value.callbacks);
+        stdx::for_each(
+            [n = ++count](auto &index) { index.collect_defaults(n); }, indices);
+        return indices;
+    }
+};
+
+} // namespace msg
--- a/include/msg/detail/rle_codec.hpp
+++ b/include/msg/detail/rle_codec.hpp
@@ -1,0 +1,97 @@
+#pragma once
+#include <lookup/entry.hpp>
+#include <lookup/input.hpp>
+
+#include <stdx/bitset.hpp>
+#include <stdx/compiler.hpp>
+#include <stdx/cx_map.hpp>
+#include <stdx/cx_vector.hpp>
+#include <stdx/tuple.hpp>
+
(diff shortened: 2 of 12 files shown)


include/msg/detail/rle_bitset.hpp
```diff
@@ -0,0 +1,191 @@
...
+namespace msg::detail {
+
+template <typename T, std::size_t NBits, typename StorageElem>
+concept ProvidesBitSets = requires(T t) {
+    requires std::invocable<T>;
+    requires std::input_iterator<decltype(std::begin(t()))>;
+    { *t().begin() } -> std::convertible_to<stdx::bitset<NBits, StorageElem>>;
+};
+
+// select a minimum sized type for indexing into the RLE data blob
```
elbeno, line 19: **thought**: stdx bitset uses almost the same function - would there be value in making it part of the stdx public API?

include/msg/detail/rle_bitset.hpp
```diff
@@ -0,0 +1,191 @@
...
+            auto num_bits = static_cast<std::size_t>(data[read_index++]);
+            if (bit_val) {
+                while (num_bits-- > 0) {
+                    result.set(bit++);
+                }
+            } else {
+                bit += num_bits;
+            }
+            bit_val = !bit_val;
+        }
```
elbeno, line 75: **suggestion**: Break out this, and the corresponding part of the make function, as functions `rle_decode` and `rle_encode`?

include/msg/detail/rle_bitset.hpp
```diff
@@ -0,0 +1,191 @@
...
+  public:
+    constexpr static auto bitset_bits = NumBitsInBitset;
+    constexpr static auto num_bitsets = NumBitSets;
+
+    using rle_data_t = std::array<std::byte, NumRLEBytes>;
+    using start_index_t =
+        typename index_types<NumBitSets, NumRLEBytes>::start_index_t;
+    using bitset_t = stdx::bitset<NumBitsInBitset, BitSetStorageElem>;
+
+    constexpr rle_multi_bitset(rle_data_t rleData, start_index_t startIndices)
```
elbeno, line 49: **suggestion**: Pass by `const &`. (Also other places). In most cases this is happening at compile time, but I prefer to be conventional anyway.

elbeno (commented):
Nice. I will let @lukevalenty also give thoughts - I know he originally thought about storage as purely external to the bitsets, but I think with the multi-bitset approach that might be a moot point. Anyway, just a few notes for now (FYI I try to use conventionalcomments.org)

include/msg/detail/rle_bitset.hpp
```diff
@@ -0,0 +1,191 @@
...
+#include <stdx/cx_vector.hpp>
+
+#include <algorithm>
+#include <numeric>
+
+namespace msg::detail {
+
+template <typename T, std::size_t NBits, typename StorageElem>
+concept ProvidesBitSets = requires(T t) {
+    requires std::invocable<T>;
```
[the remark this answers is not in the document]
elbeno: Weird, somehow this got separated from my roll-up review. Must have hit the wrong button...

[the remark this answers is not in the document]
elbeno: Btw, if you hit issues where libc++ 14 does not implement something (because libc++ has been slow to implement some C++20 features), that would be a good candidate to add to stdx.

include/msg/detail/rle_bitset.hpp
```diff
@@ -0,0 +1,191 @@
...
+    } else if constexpr (N <= std::numeric_limits<std::uint32_t>::max()) {
+        return std::uint32_t{};
+    } else {
+        return std::size_t{};
+    }
+}
+
+template <std::size_t NumBitSets, std::size_t NumRLEBytes> struct index_types {
+    using index_storage_t = decltype(select_index_storage<NumRLEBytes>());
+    using start_index_t = std::array<index_storage_t, NumBitSets>;
```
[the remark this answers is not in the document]
elbeno: Do you _really_ like it, or do you _rle_ like it?

include/msg/detail/rle_bitset.hpp
```diff
@@ -0,0 +1,191 @@
...
+    constexpr rle_multi_bitset(rle_data_t rleData, start_index_t startIndices)
+        : data{rleData}, indices{startIndices} {}
+
+    // the number of bitsets in the set
+    [[nodiscard]] constexpr auto size() const -> std::size_t {
+        return indices.size();
+    }
+
+    // get a bitset by index
+    constexpr auto get(std::size_t index) const -> bitset_t {
```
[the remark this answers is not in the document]
AshleyRoll: yeah, 100%, I mentioned in the PR description that this stuff was still to come. So will be working on that next. 😄

include/msg/detail/rle_bitset.hpp
AshleyRoll: yeah, I stole the code from there, but wanted to keep them separate until its all working, then we can find those common things and refactor.

include/msg/detail/rle_bitset.hpp
AshleyRoll: I'll have a look at this when I implement the proxy object for retrieving the bitset data. Good idea.

include/msg/detail/rle_bitset.hpp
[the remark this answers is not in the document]
AshleyRoll: added the header.

include/msg/detail/rle_bitset.hpp
AshleyRoll: fixed.

include/msg/detail/rle_bitset.hpp
AshleyRoll: done

include/msg/detail/rle_bitset.hpp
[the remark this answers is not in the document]
AshleyRoll: The proxy object is the `rle_index`. This however does not yet implement the efficient bitwise operations we want so we can avoid instantiating all the `bitset`s.

AshleyRoll:
This should now be successfully building on all compilers. The only outstanding issue is that I haven't built the more efficient bitwise operators for combining index bitsets and still materialise them for each index.

test/msg/detail/rle_codec.cpp
```diff
@@ -0,0 +1,83 @@
...
+// Encoding starts with number of zeros (0-255)
+// followed by number of ones (0-255)
+// and repeats.
+
+TEST_CASE("rle_codec can encode all zeros", "[rle_codec]") {
+    using bs = stdx::bitset<10>;
+    using codec = rle_codec<bs>;
+
+    REQUIRE_THAT(codec::encode(bs{}),
+                 RangeEquals(std::to_array<std::byte>({10_b})));
```
elbeno, line 28: **suggestion**: I have been preferring CHECK_* rather than REQUIRE_* - using REQUIRE_* only where the test must exit if the condition doesn't hold (for example typically we might REQUIRE a container has a size before CHECKing the value at an index). I don't think we are completely consistent everywhere yet, but that's what I've been trying to get to.

include/msg/detail/indexed_builder_common.hpp
```diff
@@ -0,0 +1,148 @@
...
+    constexpr auto t = BuilderValue::value.callbacks;
+    [&]<std::size_t... Is>(std::index_sequence<Is...>) -> void {
+        (fn(t[stdx::index<Is>], Is), ...);
+    }(std::make_index_sequence<t.size()>{});
+}
+
+CONSTEVAL auto with_field_index(auto t) {
+    return stdx::transform<get_field_type>([](auto x, auto) { return x; }, t,
+                                           t);
+}
```
elbeno, line 35: **suggestion**: I think this function is doing `stdx::apply_indices<get_field_type>(t)` ?

elbeno: I guess this is not new code; I just didn't notice it until now.

include/msg/detail/rle_codec.hpp
```diff
@@ -0,0 +1,96 @@
...
+                    count = 1;
+                }
+            }
+        }
+        // final byte
+        data.push_back(static_cast<std::byte>(count));
+        return data;
+    }
+
+    constexpr static auto decode(std::byte const *rle_data) -> bitset_type {
```
elbeno, line 70: **thought**: I'm accumulating use cases for ranges... I want to think of this function as transforming a stride_view over a scan... (but also, we don't yet have a scan_view)

include/msg/detail/rle_codec.hpp
```diff
@@ -0,0 +1,96 @@
...
+
+    using type = decltype(select_index_storage());
+};
+
+template <std::size_t N>
+using smallest_storage_type = typename smallest_storage<N>::type;
+
+template <typename BitSetType> struct rle_codec {
+    using bitset_type = BitSetType;
+    constexpr static auto num_bits = BitSetType{}.size();
```
elbeno, line 41: **thought**: That reminds me, https://github.com/intel/cpp-std-extensions/pull/22

include/msg/detail/indexed_builder_common.hpp
AshleyRoll: I've removed this method and replaced it with `stdx::apply_indices<>()` as you suggested to clean up.

test/msg/detail/rle_codec.cpp
AshleyRoll: Certainly. I've updated the test, and also realised I had not properly merged the last changes and this test was missing from the `CMakeLists.txt`.

include/msg/detail/rle_codec.hpp
AshleyRoll: I updated the `cpp-std-extensions` reference and modified the access to `size()` above.

include/msg/rle_indexed_builder.hpp
```diff
@@ -0,0 +1,397 @@
...
+        return static_cast<std::size_t>(storage_location);
+    }
+
+    // build a truncated copy of the storage accounting for the savings of
+    // reusing bitset RLE patterns
+    template <std::size_t FinalSize>
+    constexpr auto truncate() const -> std::array<std::byte, FinalSize> {
+        std::array<std::byte, FinalSize> result;
+        std::copy(storage.begin(), std::next(storage.begin(), FinalSize),
+                  result.begin());
```
elbeno, line 155: **suggestion**: Use `std::copy_n`

include/msg/rle_indexed_builder.hpp
```diff
@@ -0,0 +1,397 @@
...
+
+            auto full_encoded_indices =
+                stdx::make_indexed_tuple<get_field_type>(
+                    storage.template add_index<initial_offset_type>(
+                        indices)...);
+
+            struct full_encoded_result {
+                decltype(full_encoded_indices) lookup;
+                rle_storage_type storage;
+            };
```
elbeno, line 239: **praise**: Returning a `struct` with named fields instead of a `pair`/`tuple`. Nice.

elbeno (approved):

include/msg/rle_indexed_builder.hpp
AshleyRoll: great idea - I always forget about poor `copy_n` :)

include/msg/rle_indexed_builder.hpp
AshleyRoll: yeah, it makes the downstream code a lot more readable I think.

lukevalenty (approved):
finally got through the whole PR....looks great :)

PR #382, opened 2023-09-27
Try out self-hosted runners

lukevalenty:


lukevalenty:
despite some issues remaining with self hosted runners...they are still better than relying on the github runners

bdeane-intel (approved):

PR #387, opened 2023-10-02
:construction: Forks build on GitHub hosted Runners

AshleyRoll:
When a contributor forks the repository, it is not possible to use the Intel runners.

Detect the repository owner and only use the Intel runners when the workflow is running from the Intel repository.


lukevalenty (approved):
This is great! Thanks @AshleyRoll!

PR #391, opened 2023-10-13
:art: Refactor field/message matchers to new structure

elbeno:
A little refactor to take advantage of the new match library in a more natural way, hopefully.

### DIFF
--- a/include/msg/callback.hpp
+++ b/include/msg/callback.hpp
@@ -30,7 +30,7 @@ void dispatch_single_callable(CallableT const &callable,
         using MsgType = detail::msg_type_t<decltype(callable)>;
         MsgType const msg{data};
 
-        if (msg.isValid()) {
+        if (typename MsgType::matcher_t{}(msg)) {
             callable(msg, requiredArgs...);
         }
     });
@@ -71,9 +71,9 @@ struct callback_impl<BaseMsgT, extra_callback_args<ExtraCallbackArgsT...>,
 
     [[nodiscard]] constexpr auto match_any_callback() const {
         auto const matchers = stdx::transform(
-            [&](auto callback) {
-                using MsgType = detail::msg_type_t<decltype(callback)>;
-                return is_valid_msg<MsgType>(match::always);
+            [&]<typename CB>(CB const &) {
+                using MsgType = detail::msg_type_t<CB>;
+                return typename MsgType::matcher_t{};
             },
             callbacks);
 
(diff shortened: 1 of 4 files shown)


include/msg/message.hpp
```diff
@@ -66,10 +58,9 @@ struct message_base : public message_data<MaxNumDWords> {
     static_assert((... and (FieldsT::MaxDWordExtent < MaxNumDWords)));
     using FieldTupleType = stdx::tuple<FieldsT...>;
 
-    template <typename additional_matcherType>
-    [[nodiscard]] constexpr static auto match(additional_matcherType) {
-        return is_valid_msg_t<message_base, additional_matcherType>{};
-    }
+    template <typename Msg>
```
lukevalenty, line 61: question: why is this a template? why not use `message_base` instead of template argument `Msg`?

include/msg/message.hpp
```diff
@@ -66,10 +58,9 @@ struct message_base : public message_data<MaxNumDWords> {
...
+    using matcher_t = decltype(match::all(
```
lukevalenty, line 62: suggestion: a corresponding `constexpr static auto matcher_v` would be nice to have.

include/msg/callback.hpp
```diff
@@ -30,7 +30,7 @@ void dispatch_single_callable(CallableT const &callable,
         using MsgType = detail::msg_type_t<decltype(callable)>;
         MsgType const msg{data};
 
-        if (msg.isValid()) {
+        if (typename MsgType::template matcher_t<MsgType>{}(msg)) {
```
lukevalenty, line 33: suggestion: with the comments i added in message.hpp, could this look like the following?

```c++
if (MsgType::matcher_v(msg)) {
...
```

include/msg/message.hpp
elbeno: For some reason when I was writing it I thought it needed to be a template, but it doesn't.

include/msg/message.hpp
elbeno: I'd rather make a value just when it's needed?

include/msg/callback.hpp
elbeno: Without the template, this looks fine I think now even without the `matcher_v`

lukevalenty (approved):

include/msg/message.hpp
elbeno: This actually touches on a common nit that is here too: when we pass things like this as template args we are almost always constraining them to be default constructible. The matcher is passed in to the field as a template arg rather than a constructor arg. That means the matcher can't be stateful...

This might be fine but it's not enforced properly at the moment.

lukevalenty: that's true...what's the right way to enforce this?

PR #392, opened 2023-10-16
:art: Refactor msg callback

elbeno:
Collapse matchers into one instead of computing it in 3 places.

### DIFF
--- a/include/msg/callback.hpp
+++ b/include/msg/callback.hpp
@@ -12,23 +12,20 @@
 #include <utility>
 
 namespace msg {
-
-template <typename CallableT, typename DataIterableT,
-          typename... ExtraCallbackArgsT>
-void dispatch_single_callable(CallableT const &callable,
-                              DataIterableT const &data,
-                              ExtraCallbackArgsT const &...args) {
+namespace detail {
+template <typename Callable>
+void dispatch_single_callable(Callable const &callable, auto const &base_msg,
+                              auto const &...args) {
     auto const provided_args_tuple = stdx::make_tuple(args...);
     auto const required_args_tuple = stdx::transform(
-        [&](auto requiredArg) {
-            using RequiredArgType = decltype(requiredArg);
-            return get<RequiredArgType>(provided_args_tuple);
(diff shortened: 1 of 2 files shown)


include/msg/callback.hpp
```diff
@@ -12,113 +12,95 @@
...
+    constexpr static Name name{};
 
-    MatchMsgTypeT match_msg;
-    stdx::tuple<CallableTypesT...> callbacks;
+    [[nodiscard]] constexpr static auto match_any_callback() {
+        return match::any(
+            typename detail::msg_type_t<Callables>::matcher_t{}...);
+    }
+
+    decltype(std::declval<M>() and match_any_callback()) matcher;
```
elbeno, line 60: I took the opportunity to clean up some minor things, but this is the real change: we're able to form the matcher in one place.

include/msg/handler.hpp
```diff
@@ -7,21 +7,19 @@
...
+    constexpr explicit handler(Callbacks new_callbacks)
         : callbacks{new_callbacks} {}
 
-    auto is_match(BaseMsgT const &msg) const -> bool final {
-        return callbacks.fold_left(false, [&](bool state, auto &callback) {
-            return state || callback.is_match(msg);
-        });
+    auto is_match(BaseMsg const &msg) const -> bool final {
+        return stdx::any_of(
+            [&](auto &callback) { return callback.is_match(msg); }, callbacks);
```
lukevalenty, line 19: suggestion: can we wrap this in the appropriate matcher and simplify it?

include/msg/callback.hpp
lukevalenty, line 60: question: will the expression get simplified automatically?

elbeno: Yes: the boolean operator overloads call simplify on their results.

include/msg/handler.hpp
elbeno: Hm, we could have `callback` (or indeed `handler`) model `matcher`. I'm not sure we're quite there yet - maybe in a followup it will be clearer.

PR #402, opened 2023-10-19
Simplification of matchers by warranting implication rules

elbeno:


lukevalenty (approved):
I see you are ready for Walter's implication operator. 😂

PR #406, opened 2023-10-24
:art: :white_check_mark: Refactor `field_matcher`s, add relationships

elbeno:


include/msg/field_matchers.hpp
```diff
@@ -92,6 +93,111 @@ constexpr inline class remove_terms_t {
...
+template <typename FieldType, typename T, T X, T Y>
+[[nodiscard]] constexpr auto
+tag_invoke(match::implies_t, less_than_or_equal_to_t<FieldType, T, X> const &,
+           less_than_t<FieldType, T, Y> const &) -> bool {
+    return X < Y;
+}
+
+template <typename FieldType, typename T, T X, T Y>
+[[nodiscard]] constexpr auto
+tag_invoke(match::implies_t, less_than_t<FieldType, T, X> const &,
```
lukevalenty, line 181: suggestion: require integral types

elbeno: That requirement should probably be on the field itself, but I changed this to be generic. We can require incrementable when we add constraints.

mjcaisse-intel (approved):

PR #408, opened 2023-10-26
:art: Make field a view type

elbeno:
- Unify field and disjoint field
- Add field-name-based message initialization
- Change fields to be view types
- Add strong types for field locator specs


include/msg/field.hpp
```diff
@@ -22,9 +26,21 @@ concept field_spec =
     };
 
 template <typename T>
-concept field_value = field_spec<T> and requires(T t) {
-    { t.value } -> std::same_as<typename T::type &>;
-};
+concept bits_extractor =
+    std::unsigned_integral<decltype(T::size)> and
+    requires(std::span<std::uint32_t const> extract_buffer) {
+        { T::extract(extract_buffer) } -> std::same_as<std::uint64_t>;
```
mjcaisse-intel, line 32: question: why `uint64_t` here and below? For machines with native sizes smaller that 64-bits this can be an issue. Do we maybe want the native word size?

elbeno: The existing code was working in terms of `std::uint64_t` but yes, I agree. I have a couple of followups to do here; this is just the first step.

include/msg/field.hpp
```diff
@@ -88,153 +126,240 @@ template <std::uint32_t DWordIndex, std::uint32_t Lsb> struct field_locator_t {
...
+CONSTEVAL auto operator""_msb(unsigned long long int v) -> msb_t {
+    return static_cast<msb_t>(v);
+}
+// NOLINTNEXTLINE(google-runtime-int)
+CONSTEVAL auto operator""_lsb(unsigned long long int v) -> lsb_t {
+    return static_cast<lsb_t>(v);
+}
+
+struct at {
+    dword_index_t index_{};
```
mjcaisse-intel, line 217: praise: nice use of strong types

include/msg/field.hpp
```diff
@@ -88,153 +126,240 @@ template <std::uint32_t DWordIndex, std::uint32_t Lsb> struct field_locator_t {
...
+
+template <at... Ats> constexpr inline auto field_size = (0u + ... + Ats.size());
+
+template <at... Ats>
+constexpr inline auto max_dword_extent = std::max({0u, Ats.dword_extent()...});
+} // namespace detail
+
+template <typename Name, typename T = std::uint32_t, T DefaultValue = T{},
+          match::matcher M = match::always_t, at... Ats>
+class field_t : public field_spec_t<Name, T, detail::field_size<Ats...>>,
```
mjcaisse-intel, line 254: question: should `field_t` be in the detail namespace?

mjcaisse-intel (approved):
_Love_ this PR! Nicely done!

include/msg/field.hpp
elbeno: I wondered about this too. Possibly - can move it in a followup.

PR #409, opened 2023-10-27
Binary at for field locators

elbeno:
A followup to #408 - field locators don't have to specify a dword index, they can just use msb/lsb.

### DIFF
--- a/include/msg/field.hpp
+++ b/include/msg/field.hpp
@@ -8,6 +8,7 @@
 #include <stdx/ct_string.hpp>
 #include <stdx/type_traits.hpp>
 
+#include <climits>
 #include <concepts>
 #include <cstdint>
 #include <span>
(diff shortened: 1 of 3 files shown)


elbeno:
@mjcaisse-intel :tada:

include/msg/field.hpp
```diff
@@ -229,6 +227,30 @@ struct at {
...
+    [[nodiscard]] constexpr auto size() const -> std::underlying_type_t<lsb_t> {
+        return msb() - lsb() + 1;
+    }
+    [[nodiscard]] constexpr auto dword_extent() const
+        -> std::underlying_type_t<dword_index_t> {
+        return index() + (msb() / 32u);
+    }
+};
+
+template <> struct at<msb_t, lsb_t> {
```
mjcaisse-intel, line 239: issue: I think the current implementation needs to implicitly enforce the constraints with `static_assert` and nice messages or maybe the constraints can be lifted. For example, when the `msb_t` and `lsb_t` span different "dwords" things break.

```c++
field<"TestFieldNoIndex", std::uint32_t>::located<at{43_msb, 22_lsb>
```

and

```c++
field<"TestFieldNoIndex", std::uint32_t>::located<at{53_msb, 22_lsb>
```

produce cryptic (to user) errors or wrong results.

mjcaisse-intel (approved):

PR #412, opened 2023-10-31
:art: :recycle: Refactor field extraction/insertion

elbeno:
Field insertion and extraction no longer relies on the value_type in the underlying storage. The usual use case of `std::array<std::uint32_t, N>` is still the norm, but ranges over any unsigned integral type are supported.


include/msg/message.hpp
```diff
@@ -30,8 +24,8 @@ concept convertible_range_of =
     range<T> and std::convertible_to<std::iter_value_t<iterator_t<T>>, V>;
 } // namespace detail
 
-template <std::uint32_t MaxNumDWords>
-using message_data = stdx::cx_vector<std::uint32_t, MaxNumDWords>;
+template <std::uint32_t Capacity>
+using message_data = stdx::cx_vector<std::uint32_t, Capacity>;
```
mjcaisse-intel, line 28: **praise**: So much better! Expunge the term dword from all code!  (o;

mjcaisse-intel (approved):
**so-much-praise** !! Love this PR!

PR #420, opened 2023-11-10
:art: Tweaks to interrupt test

elbeno:
It's not a callback, it's a mock. And it's not an IrqImpl, it's the test HAL.


lukevalenty (approved):
Yes and thank you!

Issue #278, opened 2023-03-08
👷‍♀️ Move `irq_flow` building outside of the interrupt manager 


lukevalenty:
Right now the interrupt manager builds it's own `irq_flow` services. This is complicated and duplicates effort in `cib::nexus`. It also couples the internal details of `cib` services to the implementation of interrupt manager.

This change will remove all `irq_flow` building from the interrupt manager. It will no longer be responsible for building, initializing, or extending its `irq_flow`s. It will also remove the requirement from interrupt manager to support extending the `irq_flow`s.

Instead, the interrupt manager will use regular flows and actions for registering interrupt handlers. 

In order for the interrupt manager to determine whether an interrupt is active or not, it will need access to the nexus.

elbeno:
Resolved with #427

PR #425, opened 2023-11-15
Decouple graph builder

elbeno:
- use `tag_invoke` to walk the flow node tree; this will enable using other types in the flow tree if needed (thinking of #278)
- remove CRTP from graph_builder; it's not needed
- convert flow tests to use the "builder pattern" - in general, `add` functions that can return different types should be marked `[[nodiscard]]`? (Not yet true for `graph_builder`, but moving that way.)


lukevalenty (commented):
I think I'm going to need a walkthrough of this tomorrow.

lukevalenty (approved):

PR #431, opened 2023-11-29
:books: Add documentation for flows and interrupts

elbeno:


docs/flows.adoc
```diff
@@ -0,0 +1,404 @@
...
+
+See code at
+https://github.com/intel/compile-time-init-build/tree/main/include/flow.
+Everything in the flow library is in the `flow` namespace.
+
+A flow is a series of actions to be executed in order based on their
+dependencies. The flow library was originally designed to be used for power
+flows, moving from one power state to another. It can also be used as a generic
+extension point to register callbacks at compile-time. Flow implements the
+`constexpr` init()/build() pattern.
```
lukevalenty, line 12: suggestion: maybe we should explicitly call it a cib service...or say it implements the cib service concept

lukevalenty (approved):

docs/flows.adoc
elbeno: Sure. This text is pretty much from the existing README :)

PR #433, opened 2023-11-29
:art: Separate flow graph from builder

elbeno:
This allows us to plug in alternative graph builders, like one that can output graphviz for debugging flows.


lukevalenty (approved):
I didn't actually expect to see a graphviz builder, awesome!

Issue #446, opened 2023-12-14
Messages support for uint8_t


Thurnheer:
It would be nice if the messages would support the creation of owning uint8_t messages.
As for example Uart communication is normaly not a multiple of dword, it would be nice to have support for creation (owning) of uint8_t based messages.

lukevalenty:
Hey @bdeane-intel, does the message refactor support multiple underlying storage widths?

Thurnheer:
@lukevalenty it seems to do. I would have a PR.

elbeno:
Yep, already supported. I just added a test to demonstrate - customize the storage type passed to `owner_t`. See #315 and #448.

Issue #315, opened 2023-06-28
Accept non-narrowing conversions in the `msg::message_base` constructor


herculanodavi:
In the case of communicating with devices with 8-bit long registers using SPI or I2C, most HAL outputs are arrays of `std::uint8_t`, which could be copied safely to the `msg::message_data` type, `std::uint32_t`.

However, the constructor of `msg::message_base` asserts the `R::value_type` to be `std::uint32_t`, so this cast is not possible.  
```c++
template <detail::range R> explicit constexpr message_base(R const &r) {
    static_assert(std::is_same_v<typename R::value_type, std::uint32_t>);
    resize_and_overwrite(*this, [&](std::uint32_t *dest,
                                    std::size_t max_size) {
        std::copy_n(std::begin(r), std::min(std::size(r), max_size), dest);
        return NumDWordsT;
    });
}
```

Would it make sense to either:
1. Change the `detail::range` concept to add the requirement `std::convertible_to<std::uint32_t>` and remove the `static_assert`?
2. Change the `std::is_same_v` inside the `static_assert` to `std::is_convertible<typename R::value_type, std::uint32_t>`?

I'd be happy to make a PR for that. 

------ 

EDIT: 
Come to think of it, since the `detal::range` concept requires the two iterators, shouldn't the requirement then be `static_assert(std::is_same_v<std::iter_value_t<decltype(std::begin(r))>, std::uint32_t>);` even if no conversions are allowed? Then the `value_type` of the iterator is required, rather than that of the container. This would allow types like `uint32_t[N]` and `std::ranges::transform_view` to be inputs to message.

herculanodavi:
Thank you! This solves it for now, though using other types as storage would also be great. 

I see the issue is still open, so should I close it or would it be closed when the todo is done?

bdeane-intel:
Let's leave it open as a TODO. I think @mjcaisse-intel already has an implementation, so hopefully we can get that into cib soon.

PR #450, opened 2023-12-15
Add raw storage definition to support uin8_t based messages

Thurnheer:
Would be an addition to  #446, #315 and #448

### DIFF
--- a/include/msg/message.hpp
+++ b/include/msg/message.hpp
@@ -182,15 +182,19 @@ template <stdx::ct_string Name, typename... Fields> class message_access {
     }
 
     using default_value_type = std::uint32_t;
+    using raw_value_type = std::uint8_t;
 
     template <template <typename, std::size_t> typename C, typename T>
     using storage_t = C<T, storage_size<Fields...>::template in<T>>;
     using default_storage_t = storage_t<std::array, default_value_type>;
+    using raw_storage_t = storage_t<std::array, raw_value_type>;
 
     template <typename T>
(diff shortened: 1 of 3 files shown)


elbeno:
It's already trivial to create 8-bit storage as seen at https://github.com/intel/compile-time-init-build/blob/main/test/msg/message.cpp#L350 - I'm not sure what this adds?

Also a couple of notes:
- `raw_storage_t` isn't a good name: all storage is "raw".
- `raw_view` and `raw_owning` aren't good names: it's not raw on the outside, everything is still typed.

Honestly, I think what we have is all that's needed - a single storage definition on the outside (which only needs one definition and can then be used with any kind of message, owning or view) and a single using for the message type. Adding this extra stuff doesn't improve that?

Thurnheer:
Thank you for your honest reply and for your time reviewing my PR.

I'll probably not fully get your solution. I will try on Monday. 

Thanks for the fast reply, and I wish you a nice weekend.

elbeno:
Thanks for the issue and the PR! I hope we can get to a good solution here because the use case of `uint8_t` for storage is clearly an important one.

I did notice looking through the (existing) code that the types seem to be a little coupled to stdx::span and std::array right now, maybe I could improve that...

lukevalenty:
Hi @Thurnheer, @elbeno made a PR that adds a test with `uint8_t` backed storage. Take a look at how he creates a message with owning storage here: https://github.com/intel/compile-time-init-build/pull/448/files#diff-e2731b151a028a5dbf30f15cc8365556236a62f752824ada09a55864c8c73061R350

PR #452, opened 2023-12-19
Add test for custom view types

Thurnheer:


elbeno:
Thanks for adding these tests!
This PR is failing clang-format checks. Use a clang-17 toolchain and build the `fix-clang-format` target to fix it up.

If you are on Debian or Ubuntu, you can install a clang-17 toolchain using the llvm.sh script available at https://apt.llvm.org. Then setting the `TOOLCHAIN_ROOT` environment variable appropriately will cause cmake to pick up the right toolchain and the corresponding clang-format, e.g.

`$ TOOLCHAIN_ROOT=/usr/lib/llvm-17 cmake --preset=clang --fresh`

Alternatively, installing clang-format with pip is a good way to go if you want a systemwide installation.

Either way, the CI for cib expects the code to be formatted with the clang-format matching the latest toolchain supported, which at the moment is clang 17.

elbeno (approved):

Issue #456, opened 2023-12-20
💡`indexed_handler`: Coalesce two or more callbacks with equivalent indexed field conditions into one callback


lukevalenty:
I have a few use cases where the indexed fields for two or more callbacks will have equivalent conditions on those fields. Instead of allocating 2 or more entries in the callback array, we can combine the callbacks into a single callback with a single entry in the array. This can also give the compiler some additional opportunities to optimize the resulting assembly.

PR #459, opened 2024-01-05
:sparkles: Equivalence between matchers

elbeno:
A building block for #456. Two matchers are three-way comparable, yielding a partial ordering.

X => Y means that X > Y (intuition: "greater than" means "more strongly constrained match").

X => Y and Y => X means that X and Y are equivalent.


lukevalenty (approved):
Beautiful application of implication!

PR #461, opened 2024-01-08
:books: Update docs for msg callback usage

elbeno:


docs/message.adoc
```diff
@@ -153,11 +153,10 @@ messages; when a matcher successfully matches, the callback(s) registered will b
...
 
-// define a callback with a matcher that always matches
-constexpr auto my_callback = msg::callback<my_message>(
-    "my callback"_sc,
-    match::always,
-    [](my_message const &) { /* do something */ });
+// define a callback with the matcher from the message definition
+constexpr auto my_callback = msg::callback<"my_callback">(
+    typename my_message_defn::matcher_t{},
+    [](msg::const_view<my_message_defn>) { /* do something */ });
```
lukevalenty, line 159: What if this is a different message type?

docs/message.adoc
```diff
@@ -217,12 +219,10 @@ using my_indices = msg::index_spec<my_field>;
...
 struct my_indexed_service : msg::indexed_service<my_indices, my_message> {};
 
-// the callback is an indexed_callback
-// and here we specify the matcher
-constexpr auto my_callback = msg::indexed_callback(
-    "my indexed callback"_sc,
-    msg::equal_to<0x80>,
-    [](my_message const &) { /* do something */ });
+// this time, the callback is an indexed_callback
+constexpr auto my_callback = msg::indexed_callback<"my_indexed_callback">(
```
lukevalenty, line 223: Do we still need indexed_callback, or is it the same as a regular callback?

elbeno: I can look at that. We might be able to remove the type. But this doc is still OK because `indexed_callback` is a function - it converts the incoming matcher to sum-of-products form. But after that it might be able to make a regular callback.

lukevalenty (approved):

PR #462, opened 2024-01-09
Fix matcher ordering

elbeno:
Further research indicates that we had it the wrong way around.


lukevalenty (approved):

lukevalenty:
Does this mean `true > false` or `false > true`?

elbeno:
`true > false`

PR #469, opened 2024-01-12
Add func_decl to conveniently forward declare extension impls

lukevalenty:


mjcaisse-intel (approved):
Looking forward to seeing how the next refactor affects size and inlining.

PR #474, opened 2024-01-24
:art: Remove unnecessary type specifier for field matcher

elbeno:
The field spec already knows its type, so a matcher doesn't need to be told both the field spec and the type.


lukevalenty (approved):
It's a good ergonomic improvement! Thank you!

PR #477, opened 2024-01-28
simplify msg callback matcher and remove argument deduction

lukevalenty:
A couple changes to msg handlers and callbacks:

1. msg handler msg type is just storage.
2. msg callback template parameters now includes the msg definition after the name of the callback.
    - `msg_defn::where(...)` is no longer needed and its functionality has been pulled into `callback`
4. the callback function does not have to specify the type of the msg, it can be `auto`.
5. the callback function will always be passed all of the extra arguments. they are no longer deduced. they can be `auto`.


include/msg/detail/func_traits.hpp
elbeno, line 1: **praise**: I am glad to get rid of this!

elbeno (approved):

PR #485, opened 2024-01-31
:sparkles: Add alternative callback matcher syntax

elbeno:


elbeno:
Opened as draft - let me know if you think we can improve the syntax.

lukevalenty (approved):

PR #486, opened 2024-01-31
ensure message views are sent to callbacks on indexed handlers

lukevalenty:


include/lookup/detail/select.hpp
```diff
@@ -22,79 +22,48 @@ constexpr inline auto fallback_select_lt(K lhs, K rhs, T first, T second) -> T {
...
-                                    T first, T second) -> T {
-    if constexpr (sizeof(T) <= 4) {
-        T result = second;
+template <typename K, typename T>
+    requires(sizeof(T) <= 4) and (sizeof(K) <= 4)
+static inline auto optimized_select(K raw_lhs, K raw_rhs, T first, T second)
+    -> T {
+    T result = second;
+    auto const lhs = static_cast<std::uint32_t>(raw_lhs);
+    auto const rhs = static_cast<std::uint32_t>(raw_rhs);
```
elbeno, line 31: **question**: Do we want `stdx::bit_cast` here? (and lines 51-52)?

lukevalenty: yes, i think we do

lukevalenty: ehh, no, we can't do that. the sizes may be different and that's a requirement for bit_cast. 

https://en.cppreference.com/w/cpp/numeric/bit_cast

```c++
template<class To, class From>
[std::enable_if_t](http://en.cppreference.com/w/cpp/types/enable_if)<
    sizeof(To) == sizeof(From) &&
    [std::is_trivially_copyable_v](http://en.cppreference.com/w/cpp/types/is_trivially_copyable)<From> &&
    [std::is_trivially_copyable_v](http://en.cppreference.com/w/cpp/types/is_trivially_copyable)<To>,
    To>
// constexpr support needs compiler magic
bit_cast(const From& src) noexcept;
```

PR #487, opened 2024-01-31
:sparkles: Add alt syntax for `in` field matcher

elbeno:


lukevalenty (approved):
I like that syntax very much

PR #491, opened 2024-02-05
use `to_string` function to convert field values into string_constants

lukevalenty:
`msg::field::describe` will now call `to_string(value_type)` to convert the field's value into a string_constant. This allows for values to be described even though they can not be directly converted into `std::uint32_t`. With this new extension point, the ability to correctly print 64-bit unsigned values is added.

### DIFF
--- a/include/msg/field.hpp
+++ b/include/msg/field.hpp
@@ -371,6 +371,21 @@ template <typename R> constexpr auto capacity() -> std::size_t {
     }
 }
 
+[[nodiscard]] constexpr static auto to_string(std::integral auto v) {
+    constexpr auto hex_digits = sizeof(v) * 2;
+
+    if constexpr (hex_digits <= 8) {
+        constexpr auto hex_digits_sc = format("{}"_sc, sc::uint_<hex_digits>);
+        constexpr auto fmt_sc = "0x{:0"_sc + hex_digits_sc + "x}"_sc;
+        return format(fmt_sc, v);
+
+    } else {
+        return format("0x{:08x}{:08x}"_sc,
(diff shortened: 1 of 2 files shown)


include/msg/field.hpp
```diff
@@ -371,6 +371,21 @@ template <typename R> constexpr auto capacity() -> std::size_t {
...
+
+    if constexpr (hex_digits <= 8) {
+        constexpr auto hex_digits_sc = format("{}"_sc, sc::uint_<hex_digits>);
+        constexpr auto fmt_sc = "0x{:0"_sc + hex_digits_sc + "x}"_sc;
+        return format(fmt_sc, v);
+
+    } else {
+        return format("0x{:08x}{:08x}"_sc,
+                      static_cast<std::uint32_t>(v >> 32ull),
+                      static_cast<std::uint32_t>(v & 0xffffffff));
```
elbeno, line 385: **question**: What use is this code path for?

elbeno: 64-bit values use the other code path though... I was wondering which use case came through here?

elbeno: Oh... I see. yuck. We're coupling to mipi protocol here? :(

lukevalenty: Goooood point. Not mipi protocol, but some other implementation detail that doesn't belong here. I will remove this branch and request review tomorrow.

elbeno: We have all the types in the generated json, so can we figure this out in the mipi logger? Rather than imposing on the call sites?

PR #494, opened 2024-02-06
:sparkles: Add capabilities to gen_str_catalog.py

elbeno:
- Properly handle multiple input libraries
- Allow extra input JSON to be copied to output


lukevalenty:
Just took a quick look. Awesome! We should use it! 😂

lukevalenty (approved):

PR #495, opened 2024-02-07
:sparkles: Support extra headers for generated strings TU

elbeno:
For example, in order to format enums at runtime, the types need to be defined.


lukevalenty (approved):
👍🏻

PR #499, opened 2024-02-10
:sparkles: Add module ID to logs

elbeno:
A module ID is a compile-time string. Module ID is overridden in the current scope with the `CIB_LOG_MODULE` macro. Module IDs are mapped in the same way as string catalog IDs, with a link-time dependency.


lukevalenty (approved):
praise: I really like this interface...will be a nice and easy way to add log modules without being too verbose in each log statement.

PR #504, opened 2024-02-16
:safety_vest: Guard against improper construction of owning messages

elbeno:


lukevalenty (approved):

include/msg/message.hpp
```diff
@@ -345,6 +345,16 @@ template <stdx::ct_string Name, typename... Fields> struct message {
 
         template <detail::storage_like S, field_value... Vs>
         explicit constexpr owner_t(S const &s, Vs... vs) {
+            static_assert(std::is_same_v<typename S::value_type,
+                                         typename storage_t::value_type>,
+                          "Attempted to construct owning message with "
+                          "incompatible storage");
+
+            static_assert(stdx::ct_capacity_v<S> <=
```
elbeno, line 353: **thought**: Actually, now we have ct_capacity, we can compute the max to copy at compile time, and safety construct messages from over/undersized storage. What do you think @lukevalenty ?

elbeno: Because it really ought to be possible to do:
```cpp
std::array<std::uint32_t, 4> a{};
owning<msg_defn> m{a};  // where the default size of the defn is less than the array size
```

include/msg/message.hpp
```diff
@@ -345,7 +345,14 @@ template <stdx::ct_string Name, typename... Fields> struct message {
 
         template <detail::storage_like S, field_value... Vs>
         explicit constexpr owner_t(S const &s, Vs... vs) {
-            std::copy(std::begin(s), std::end(s), std::begin(storage));
+            static_assert(std::is_same_v<typename S::value_type,
+                                         typename storage_t::value_type>,
+                          "Attempted to construct owning message with "
+                          "incompatible storage");
+
+            constexpr auto N = std::max(stdx::ct_capacity_v<S>,
```
lukevalenty, line 353: issue: shouldn't this be `std::min`?

include/msg/message.hpp
lukevalenty: i think it definitely makes sense to construct a message from a larger buffer...but constructing a message from a smaller buffer could be an error. i don't think we should implicitly allow that.

include/msg/message.hpp
elbeno: Yes it should!

include/msg/message.hpp
elbeno: I can imagine the use case for either though: either having a storage area that can fit one of a number of messages, then copying a small message out, or making a message header and then copying it into an owning message.

elbeno: Neither is implicit: all these constructors are explicit because they involve copying data.

elbeno:
This completes #501

PR #512, opened 2024-03-14
:books: Add docs for logging and sc libraries

bdeane-intel:


docs/logging.adoc
```diff
@@ -0,0 +1,174 @@
...
+    ERROR = 2,
+    WARN = 3,
+    INFO = 4,
+    USER1 = 5,
+    USER2 = 6,
+    TRACE = 7
+};
+----
+
+NOTE: For backward compatibility, this is an unscoped enumeration.
```
lukevalenty, line 38: Question: I wonder if it would be much of an issue to make this scoped.

docs/logging.adoc
```diff
@@ -0,0 +1,174 @@
...
+    }
+  } logger;
+};
+----
+
+Notice that the first two template parameters to log are the
+xref:logging.adoc#_log_levels[level] and the xref:logging.adoc#_modules[module
+ID]. The `ModuleId` type is a xref:sc.adoc#_string_constants[compile-time string].
+
+The first two runtime parameters receive preprocessor `\_​_FILE_​\_` and `__LINE_​_` values
```
lukevalenty, line 105: Formatting: backslash on one but not the either...which do we need to format this correctly?

docs/sc.adoc
```diff
@@ -0,0 +1,113 @@
...
+
+// integral constants likewise
+sc::format("{} is a taxicab number"_sc, std::integral_constant<int, 1729>{}); // "1729 is a taxicab number"
+
+// and names of types
+sc::format("When in doubt, do as the {}s do"_sc, sc::type_<int>); // "When in doubt, do as the ints do"
+----
+
+When using _cib_ xref:logging.adoc#_log_macros[log macros], any arguments that
+can be formatted at compile-time are automatically interpolated into the
```
lukevalenty, line 111: Comment: you may want to clarify the requirements for compile-time values vs runtime values. For example, a constexpr int on its own will still be formatted at runtime. Wrapping it in an integral_constant is the only way the format function knows how to format it at compile time.

lukevalenty (approved):

docs/logging.adoc
bdeane-intel: Honestly, it's a mystery. I looked at the asciidoc documentation... this is what works.

docs/logging.adoc
bdeane-intel: We can try it.

PR #514, opened 2024-03-21
:art: Refactor field to use stdx::bit_mask

bdeane-intel:
Simplified `bits_locator_t::insert` a little: two compile-time cases can be collapsed into one.

### DIFF
--- a/include/msg/field.hpp
+++ b/include/msg/field.hpp
@@ -72,19 +72,6 @@ concept field_location = requires(T const &t) {
 };
 
 namespace detail {
-template <typename T> CONSTEVAL auto bit_size() { return sizeof(T) * CHAR_BIT; }
-
-template <std::unsigned_integral T, auto BitSize = bit_size<T>()>
-    requires(BitSize <= bit_size<T>())
-CONSTEVAL auto bit_mask() -> T {
-    if constexpr (BitSize == 0u) {
-        return {};
-    } else {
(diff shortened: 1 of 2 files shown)


include/msg/field.hpp
```diff
@@ -171,47 +158,44 @@ struct bits_locator_t {
         constexpr auto BaseIndex =
             DWordIndex * sizeof(std::uint32_t) / sizeof(elem_t);
 
-        constexpr auto elem_size = bit_size<elem_t>();
-        constexpr auto max_idx = BaseIndex + Msb / elem_size;
-        constexpr auto min_idx = BaseIndex + Lsb / elem_size;
+        constexpr auto elem_size = stdx::bit_size<elem_t>();
+        [[maybe_unused]] constexpr auto min_idx = BaseIndex + Lsb / elem_size;
+        [[maybe_unused]] constexpr auto max_idx = BaseIndex + Msb / elem_size;
```
bdeane-intel, line 163: **note**: GCC thinks that `max_idx` is unused in this code. The odd thing is that `min_idx` could be unused if we take the branch on line 170... (and GCC doesn't complain about that) but there's no way `max_idx` is ever unused - it's used unconditionally on line 170. Anyway, I just put `[[maybe_unused]]` on both.

mjcaisse-intel (approved):

PR #519, opened 2024-03-29
:sparkles: Allow string catalog generation to use stable IDs

### LINKED ISSUE #517: String catalog should generate stable IDs
Given previous collateral (XML/JSON), string catalog generation should retain the indices of strings that were previously present.

elbeno:
Closes #517

### DIFF
--- a/tools/gen_str_catalog.py
+++ b/tools/gen_str_catalog.py
@@ -92,7 +92,11 @@ def extract(line_num, line_m):
         )
 
 
-def read_input(filenames):
+def stable_msg_key(msg):
+    return hash(msg["level"]) ^ hash(msg["msg"]) ^ hash("".join(msg["arg_types"]))
+
+
+def read_input(filenames, stable_ids):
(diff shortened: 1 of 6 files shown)


tools/gen_str_catalog.py
```diff
@@ -192,6 +201,14 @@ def write_json(messages, extra_inputs, filename):
         json.dump(str_catalog, f, indent=4)
 
 
+def read_stable(stable_filenames):
+    stable_catalog = dict()
+    for filename in stable_filenames:
+        with open(filename, "r") as f:
+            stable_catalog.update(json.load(f))
+    return {msg["msg"]: msg["id"] for msg in stable_catalog["messages"]}
```
mjcaisse-intel, line 209: concern: I believe the stable portion needs to include all parts of the message (e.g. `level`, `msg`, args ...) and not just the msg. The only portion that isn't part of the stable comparison is the `id`.

elbeno: OK, good call - I factored out the `stable_msg_key` function for this.

mjcaisse-intel (approved):

PR #527, opened 2024-04-12
:sparkles: Add `msg::view_of` and `msg::const_view_of`

elbeno:
And allow message views to be implicitly constructed from larger (typically oversized) views.


include/msg/message.hpp
```diff
@@ -326,10 +331,18 @@ template <stdx::ct_string Name, typename... Fields> struct message {
...
+                                  typename Span::element_type> and
+                     Span::extent <= S::extent)
+        // NOLINTNEXTLINE(google-explicit-constructor)
+        constexpr view_t(view_t<S> const &s) : storage{s.data()} {}
+
         [[nodiscard]] constexpr auto data() const { return storage; }
 
       private:
-        static_assert((... and Fields::template fits_inside<Span>()),
+        static_assert(definition_t::fits_inside<span_t>,
```
lukevalenty, line 345: question: why `span_t` here but `Span` on line 347?

lukevalenty: (and elsewhere within this struct)

include/msg/message.hpp
```diff
@@ -326,10 +331,18 @@ template <stdx::ct_string Name, typename... Fields> struct message {
             this->set(vs...);
         }
 
+        template <typename S>
+            requires(not std::same_as<S, Span> and
+                     std::same_as<std::add_const_t<typename S::element_type>,
+                                  typename Span::element_type> and
+                     Span::extent <= S::extent)
```
lukevalenty, line 338: question: do we need to consider dynamic extent?

lukevalenty (approved):

include/msg/message.hpp
elbeno: No reason - I've updated it to be `span_t`  to be consistent with `owner`'s use of `storage_t`

include/msg/message.hpp
elbeno: No; messages with dynamic extent aren't supported and already don't compile for reasons elsewhere. (I tried that, and decided to remove the dynamic extent code here to make it simpler.)

PR #530, opened 2024-04-17
add pseudo_pext perfect hash function to lookup

lukevalenty:


lukevalenty:
[Benchmarked](test/lookup/pseudo_pext.cpp) against [boost-ext/mph](https://github.com/boost-ext/mph):


image

krzysztof-jusiak:
Results  on my machine:

```
|               ns/op |                op/s |    err% |     total | benchmark                                          
|--------------------:|--------------------:|--------:|----------:|:----------                                         
|                4.36 |      229,284,958.02 |    0.0% |      1.04 | `small pseudo pext lookup`                         
|                3.49 |      286,435,356.88 |    0.0% |      0.84 | `small mph pext lookup`                            
```

mph generates the following x86-64 asm (https://godbolt.org/z/vfdfG7aKr)

```
main:                                   # @main
        mov     eax, 464
        pext    eax, edi, eax
        lea     rcx, [rip + _ZZNK3mph6v1_0_74pextILm7EXtlNS0_13unconditionalMUlbT_T0_E_EEEEclITnDaXtlNS0_8optionalIjLj325EEELt325EEETnDaXtlSt5arrayISt4pairIjjELm5EEtlA5_SC_tlSC_tlSt11__pair_baseIjjEELj54ELj91EEtlSC_tlSG_ELj324ELj54EEtlSC_tlSG_ELj64ELj324EEtlSC_tlSG_ELj234ELj64EEtlSC_tlSG_ELj91ELj234EEEEERKjjTnDaLj464EJEQaasr8conceptsE12bits_size_leIXTL0_3_EXT_EEsr8conceptsE16all_keys_size_ltIXTL0_0_EXstTL0_2_EEEEDtT_EOT1_DpOT4_E5index]
        movzx   eax, word ptr [rcx + 2*rax]
        ret
_ZZNK3mph6v1_0_74pextILm7EXtlNS0_13unconditionalMUlbT_T0_E_EEEEclITnDaXtlNS0_8optionalIjLj325EEELt325EEETnDaXtlSt5arrayISt4pairIjjELm5EEtlA5_SC_tlSC_tlSt11__pair_baseIjjEELj54ELj91EEtlSC_tlSG_ELj324ELj54EEtlSC_tlSG_ELj64ELj324EEtlSC_tlSG_ELj234ELj64EEtlSC_tlSG_ELj91ELj234EEEEERKjjTnDaLj464EJEQaasr8conceptsE12bits_size_leIXTL0_3_EXT_EEsr8conceptsE16all_keys_size_ltIXTL0_0_EXstTL0_2_EEEEDtT_EOT1_DpOT4_E5index:
        .short  325                             # 0x145
        .short  91                              # 0x5b
        .short  324                             # 0x144
        .short  234                             # 0xea
        .short  325                             # 0x145
        .short  325                             # 0x145
        .short  64                              # 0x40
        .short  325                             # 0x145
        .short  325                             # 0x145
        .short  325                             # 0x145
        .short  54                              # 0x36
        .short  325                             # 0x145
        .short  325                             # 0x145
        .short  325                             # 0x145
        .short  325                             # 0x145
        .short  325       
```

Though not sure whether the benchmark exercise it too much as there should be difference in mph with the following policy as it's doing less lookups

```
constexpr auto policies = []<const auto...ts>(auto&&... args) {
  return mph::pext<7u, mph::unconditional>{}.template operator()<ts...>(std::forward<decltype(args)>(args)...);
};

constexpr auto mph_map = mph::hash<
    std::array{
        std::pair{ 54u,  91u},
        std::pair{324u,  54u},
        std::pair{ 64u, 324u},
        std::pair{234u,  64u},
        std::pair{ 91u, 234u}
  }, policies
>;

static_assert(mph_map( 12u) == false);
static_assert(*mph_map( 54u) ==  91u);
static_assert(*mph_map(324u) ==  54u);
static_assert(*mph_map( 64u) == 324u);
static_assert(*mph_map(234u) ==  64u);
static_assert(*mph_map( 91u) == 234u);
```

but the results seems the same.

lukevalenty:
@krzysztof-jusiak: what happens when the policy is changed to check that the key is a valid key?

krzysztof-jusiak:
As discussed via linkedin merged the lookup and index arrays to avoid additional load - https://godbolt.org/z/9n8bbTG81 (tthat's with the valid key comparion)

```
main:                                   # @main
        mov     eax, 464
        lea     rdx, [rip + _ZZNK3mph6v1_0_74pextILm7EXtlNS0_11conditionalMUlbT_T0_E_EEEEclITnDaLj325ETnDaXtlSt5arrayISt4pairIjjELm5EEtlA5_SA_tlSA_tlSt11__pair_baseIjjEELj54ELj91EEtlSA_tlSE_ELj324ELj54EEtlSA_tlSE_ELj64ELj324EEtlSA_tlSE_ELj234ELj64EEtlSA_tlSE_ELj91ELj234EEEEERKjjTnDaLj464EJEQaasr8conceptsE12bits_size_leIXTL0_3_EXT_EEsr8conceptsE16all_keys_size_ltIXTL0_0_EXstTL0_2_EEEEDtT_EOT1_DpOT4_E11keys_values]
        pext    ecx, edi, eax
        mov     eax, 325
        cmp     dword ptr [rdx + 8*rcx], edi
        jne     .LBB0_2
        mov     eax, dword ptr [rdx + 8*rcx + 4]
.LBB0_2:
        ret
_ZZNK3mph6v1_0_74pextILm7EXtlNS0_11conditionalMUlbT_T0_E_EEEEclITnDaLj325ETnDaXtlSt5arrayISt4pairIjjELm5EEtlA5_SA_tlSA_tlSt11__pair_baseIjjEELj54ELj91EEtlSA_tlSE_ELj324ELj54EEtlSA_tlSE_ELj64ELj324EEtlSA_tlSE_ELj234ELj64EEtlSA_tlSE_ELj91ELj234EEEEERKjjTnDaLj464EJEQaasr8conceptsE12bits_size_leIXTL0_3_EXT_EEsr8conceptsE16all_keys_size_ltIXTL0_0_EXstTL0_2_EEEEDtT_EOT1_DpOT4_E11keys_values:
        .long   0                               # 0x0
        .long   325                             # 0x145
        .long   54                              # 0x36
        .long   91                              # 0x5b
        .long   64                              # 0x40
        .long   324                             # 0x144
        .long   91                              # 0x5b
        .long   234                             # 0xea
        .long   0                               # 0x0
        .long   325                             # 0x145
        .long   0                               # 0x0
        .long   325                             # 0x145
        .long   234                             # 0xea
        .long   64                              # 0x40
        .long   0                               # 0x0
        .long   325                             # 0x145
        .long   0                               # 0x0
        .long   325                             # 0x145
        .long   0                               # 0x0
        .long   325                             # 0x145
        .long   324                             # 0x144
        .long   54                              # 0x36
        .long   0                               # 0x0
        .long   325                             # 0x145
        .long   0                               # 0x0
        .long   325                             # 0x145
        .long   0                               # 0x0
        .long   325                             # 0x145
        .long   0                               # 0x0
        .long   325                             # 0x145
        .long   0                               # 0x0
        .long   325                             # 0x145
```

lukevalenty:
> As discussed via linkedin merged the lookup and index arrays to avoid additional load - https://godbolt.org/z/9n8bbTG81 (tthat's with the valid key comparion)

That's a nice speedup for `mph`! 😄

include/lookup/pseudo_pext_lookup.hpp
```diff
@@ -0,0 +1,164 @@
...
+        using value_type = typename decltype(input)::value_type;
+
+        constexpr auto mask = detail::find_pseudo_pext_mask(input.entries);
+        constexpr auto p = detail::pseudo_pext_t(mask);
+        constexpr auto storage_size = 1 << std::popcount(mask);
+        constexpr auto storage = [&]() {
+            std::array<entry<key_type, value_type>, storage_size> s{};
+
+            std::fill(s.begin(), s.end(),
+                      entry<key_type, value_type>{0, input.default_value});
```
elbeno, line 151: **suggestion**: Use `array::fill`

include/lookup/pseudo_pext_lookup.hpp
```diff
@@ -0,0 +1,164 @@
...
+}
+
+template <typename T, std::size_t S>
+CONSTEVAL std::array<T, S> with_mask(T const mask,
+                                     std::array<T, S> const &keys) {
+    std::array<T, S> new_keys{};
+
+    for (auto i = std::size_t{}; i < S; i++) {
+        new_keys[i] = pseudo_pext_t(mask)(keys[i]);
+    }
```
elbeno, line 87: **suggestion**: `std::transform`?

include/lookup/pseudo_pext_lookup.hpp
```diff
@@ -0,0 +1,164 @@
...
+    return new_keys;
+}
+
+template <typename T, typename V, std::size_t S>
+CONSTEVAL auto get_keys(std::array<entry<T, V>, S> const &pairs) {
+    std::array<T, S> new_keys{};
+
+    for (auto i = std::size_t{}; i < S; i++) {
+        new_keys[i] = pairs[i].key_;
+    }
```
elbeno, line 98: **suggestion**: `std::transform`?

include/lookup/pseudo_pext_lookup.hpp
```diff
@@ -0,0 +1,164 @@
...
+
+    constexpr pseudo_pext_t(T mask_arg) {
+        mask = mask_arg;
+        constexpr auto t_digits = std::numeric_limits<T>::digits;
+        auto const num_bits_to_extract = std::popcount(mask);
+        auto const left_padding = std::countl_zero(mask);
+        gap_bits = static_cast<std::size_t>(t_digits - num_bits_to_extract -
+                                            left_padding);
+        coefficient = compute_pack_coefficient<T>(gap_bits, mask);
+        final_mask = (T{1} << num_bits_to_extract) - T{1};
```
elbeno, line 58: **question**: Is there a way to call https://intel.github.io/cpp-std-extensions/#_bit_mask here? All of this is known at compile time... maybe we need a version of `bit_mask` that works on values as well as template arguments...

include/lookup/pseudo_pext_lookup.hpp
```diff
@@ -0,0 +1,164 @@
...
+
+        if (curr_src_bit_set) {
+            dst += 1;
+        }
+
+        prev_src_bit_set = curr_src_bit_set;
+    }
+
+    return pack_coefficient;
+}
```
elbeno, line 42: **question**: Would it perhaps be easier to compute the mask using a `bitset` and convert to integral type afterwards?

include/lookup/pseudo_pext_lookup.hpp
lukevalenty: need a version that takes values, like you mention

include/lookup/pseudo_pext_lookup.hpp
lukevalenty: might be nice...but the conversions out of bitset are only unsigned long and unsigned long long. i would need to static_cast that to the appropriate type....yeah...it's probably better to use bitset.

lukevalenty: actually, bitset isn't constexpr until C++23 😭

lukevalenty:
@elbeno I've updated the code in response to your review comments.

include/lookup/pseudo_pext_lookup.hpp
elbeno: But `stdx::bitset` is. And if you need a better conversion than just unsigned long (long), we can do that :)

include/lookup/pseudo_pext_lookup.hpp
elbeno: Need to be able to use consteval function args in templates... :)

include/lookup/pseudo_pext_lookup.hpp
lukevalenty: Ohhhh! Yes, I like that idea very much! 😄

I want something like 'to<int32_t>()' to do the conversion.

lukevalenty: using `stdx::bitset` with a static_cast which is fine for now.

krzysztof-jusiak:
I've played with a bit more and noticed a few  things.

Not surprisingly, depending on the benchmark and what the code is optimized for the results are vastly different.

For example, when optimizing for uops the following https://godbolt.org/z/71xce1EqW yields

```cpp
uOps Per Cycle:    5.44
IPC:               5.44
Block RThroughput: 1.3
```

```cpp
main:                                   # @main
        mov     eax, 464
        lea     rcx, [rip + _ZZNK3mph6v1_0_74pextILm7EXtlNS0_11conditionalMUlbT_T0_E_EEEEclITnDaXtlNS0_8optionalIjLj325EEELt325EEETnDaXtlSt5arrayISt4pairIjjELm5EEtlA5_SC_tlSC_tlSt11__pair_baseIjjEELj54ELj91EEtlSC_tlSG_ELj324ELj54EEtlSC_tlSG_ELj64ELj324EEtlSC_tlSG_ELj234ELj64EEtlSC_tlSG_ELj91ELj234EEEEERKjjTnDaLj464EJEQaasr8conceptsE12bits_size_leIXTL0_3_EXT_EEsr8conceptsE16all_keys_size_ltIXTL0_0_EXstTL0_2_EEEEDtT_EOT1_DpOT4_E6lookup]
        xor     edx, edx
        pext    eax, edi, eax
        cmp     dword ptr [rcx + 8*rax], edi
        movzx   eax, word ptr [rcx + 8*rax + 4]
        cmovne  eax, edx
        ret
_ZZNK3mph6v1_0_74pextILm7EXtlNS0_11conditionalMUlbT_T0_E_EEEEclITnDaXtlNS0_8optionalIjLj325EEELt325EEETnDaXtlSt5arrayISt4pairIjjELm5EEtlA5_SC_tlSC_tlSt11__pair_baseIjjEELj54ELj91EEtlSC_tlSG_ELj324ELj54EEtlSC_tlSG_ELj64ELj324EEtlSC_tlSG_ELj234ELj64EEtlSC_tlSG_ELj91ELj234EEEEERKjjTnDaLj464EJEQaasr8conceptsE12bits_size_leIXTL0_3_EXT_EEsr8conceptsE16all_keys_size_ltIXTL0_0_EXstTL0_2_EEEEDtT_EOT1_DpOT4_E6lookup:
        .zero   8
        .long   54                              # 0x36
        .long   91                              # 0x5b
        .long   64                              # 0x40
        .long   324                             # 0x144
        .long   91                              # 0x5b
        .long   234                             # 0xea
        .zero   8
        .zero   8
        .long   234                             # 0xea
        .long   64                              # 0x40
        .zero   8
        .zero   8
        .zero   8
        .long   324                             # 0x144
        .long   54                              # 0x36
        .zero   8
        .zero   8
        .zero   8
        .zero   8
        .zero   8
```

However, it's not the fastest in benchmarks where they key is not found that often. In such cases having 2 tables seems to yield better results as the second lea is mostly avoided.

Trying to optimize for storage (based on min/max values for key/values) is beneficial only when there are a lot of keys/values to minimize the cache usage but if the number of elements is small actually the most beneficial is to keep key and value 4/8 bytes and even convert them to the same type.

Interestingly aligning to the cache_line size it's only beneficial with string like keys and actually worse for integral keys.

Using 0 for the unknown value generates better code and perf than max+1 or -1 but that also depends on the keys as it may require sub.

Ideally if it can be guaranteed that the input is always on of the keys, unconditional (no cmp vs the key) move can be done but that's often not the case.

clang often generates cmov whereas gcc doesn't do it basically ever, however, gcc seems to be a bit faster on average on different benchmarks (especially for string like keys).

All in all, the perf would have to be measured in real use cases to be sure (maybe PGO would also be beneficial but that may overfill so hard to say). Likely keeping options open via policies is the best bet.

elbeno (approved):

PR #531, opened 2024-04-18
use stdx bit_mask and bitset for pseudo_pext algorithm

lukevalenty:


include/lookup/pseudo_pext_lookup.hpp
```diff
@@ -41,24 +42,25 @@ constexpr auto compute_pack_coefficient(std::size_t dst, T const mask) -> T {
...
 
 template <typename T> struct pseudo_pext_t {
-    T mask;
-    T coefficient;
-    T final_mask;
-    std::size_t gap_bits;
+    [[no_unique_address]] T mask;
+    [[no_unique_address]] T coefficient;
+    [[no_unique_address]] T final_mask;
+    [[no_unique_address]] std::size_t gap_bits;
```
bdeane-intel, line 52: **question**: Why? I don't think it's possible for them to be empty types? Would it be better to constrain the template param to be e.g. unsigned integral?

lukevalenty: hmmm...i'll remove these. i had a concern about the compiler allocating memory for this struct and i didn't want it to necessarily have to allocate everything if it could avoid it. i would prefer the compiler use immediate values in the assembly vs storing the struct in memory.

lukevalenty: i removed them

bdeane-intel (approved):

PR #552, opened 2024-05-21
:art: Rework field matchers/default values

### LINKED ISSUE #551: Add a field requirement that it has no default
For constructing messages, it would be nice to have a field without a default that requires a value be provided at construction.
So it would be a compile error to make a message without setting the port ID (for instance).

elbeno:
- Make field matcher aliases snake_case
- Preserve legacy field matcher aliases for now
- Add field default aliases
- Compile-time error when messages are constructed with uninitialized fields

Closes #551

### DIFF
--- a/docs/message.adoc
+++ b/docs/message.adoc
@@ -78,12 +78,12 @@ For example, a message type looks like this:
 ----
 using my_message_defn = msg::message<
     "my_message",                      // name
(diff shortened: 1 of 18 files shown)


bdeane-intel:
@rubyliuintelcom

mjcaisse-intel (approved):

PR #553, opened 2024-05-30
:construction_worker: Add dependency caching in CI

elbeno:


elbeno:
CMake configure step with uncached dependencies: ~45s.
CMake configure step with cached dependencies: ~5s.

Issue #561, opened 2024-06-14
refactor and clean up pseudo_pext benchmark python code


lukevalenty:
Lots of feedback to refactor the pseudo_pext benchmark code in https://github.com/intel/compile-time-init-build/pull/535, this issue is being filed to track it's completion.

elbeno:
- [ ] the pseudo_pext benchmark code needs to be moved out of the test area and into a separate benchmarking area
- [ ] benchmarking targets shouldn't be part of the ALL target
- [ ] pseudo_pext.cpp uses a C++23 feature (static variable in constexpr function); either needs to be changed for C++20 or run only with appropriate toolchains

PR #562, opened 2024-06-17
:bug: Fix up cmake for pseudo pext benchmarks

elbeno:
The pseudo pext code needs more help; I can't even build it yet (tried clang 17, 18, gcc 13).

I've started making a task list in issue https://github.com/intel/compile-time-init-build/issues/561


lukevalenty (approved):
Good call removing the pseudo_pext benchmarks from all.

PR #574, opened 2024-07-13
msg::callback::handle args determined by caller

mjcaisse-intel:
Previous versions of message callback handler construction would deduce the `ExtraArgs` from the callable passed in by the user when forming the callback. This is no longer the case which means that the signature of `msg::callback::handle` does not match the signature of the contained `callable`.

This PR works around the issue; however, the real question is, "do we want to deduce the `ExtraArgs` from the user's handler? Deducing has some real problems because the handler is not always concrete (e.g. a generic lambda). I think the PR solution is probably the correct route.

### DIFF
--- a/include/msg/callback.hpp
+++ b/include/msg/callback.hpp
@@ -20,15 +20,16 @@ namespace detail {
  * and handling incoming messages.
  */
 template <stdx::ct_string Name, typename Msg, match::matcher M,
-          stdx::callable F, typename... ExtraArgs>
+          stdx::callable F>
 struct callback {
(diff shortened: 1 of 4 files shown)


elbeno:
If we remove `ExtraArgs...` in this one place we should strip it out from everywhere; it's now unused, right?

mjcaisse:
> If we remove `ExtraArgs...` in this one place we should strip it out from everywhere; it's now unused, right?

I believe so. I'll update the PR.

rubyliuintelcom (approved):

lukevalenty (approved):

PR #584, opened 2024-07-19
:sparkles: Add capability for flavored (e.g. secure) logs

elbeno:
I decided to call them "flavoured" rather than "coloured" to avoid confusion with colour in log output.


lukevalenty (approved):
Looks great to me!

PR #588, opened 2024-07-31
:sparkles: Add message equivalence

elbeno:
Problem:
- equality is not defined for span, so it is also not defined for messages.

Solution:
- add `msg::equivalent` to determine whether two message instances are
  equivalent (all fields have the same values).
- equivalence is defined for all combinations of owning, const and mutable view
  types.

### DIFF
--- a/test/msg/message.cpp
+++ b/test/msg/message.cpp
@@ -3,6 +3,14 @@
 
 #include <catch2/catch_test_macros.hpp>
 
+#include <algorithm>
+#include <array>
+#include <cstdint>
+#include <functional>
+#include <iterator>
+#include <string>
+#include <type_traits>
+
 namespace {
 using namespace msg;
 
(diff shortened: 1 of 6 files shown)


test/msg/message.cpp
```diff
@@ -431,3 +442,80 @@ TEST_CASE("extend message with new field constraint", "[message]") {
...
+    test_uint8_msg m1{"f1"_field = 0xba11, "f2"_field = 0x42,
+                      "f3"_field = 0xd00d};
+    test_uint8_msg m2{"f1"_field = 0xba11, "f2"_field = 0x42,
+                      "f3"_field = 0xd00d};
+    // no field covers [0_dw, 23_msb, 16_lsb]
+    // i.e. byte 2 is not salient
+    m1.data()[2] = 0x11;
+    CHECK(m2.data()[2] == 0);
+
+    CHECK(m1 != m2);
```
mjcaisse-intel, line 512: It is not obvious to me that this is correct. Owning messages are morally equivalent to structs:

```c++
struct foo {
  uint8_t a;
  uint16_t b;
};

foo a{8, 10};
foo b(8, 10};

assert(a == b);
```

Padding in the structure is part of equality. Equality is not asking of a and b are the same object. For owning messages I would expect field comparison. Thoughts?

bdeane-intel: Padding _is_ part of equality, or is not?

At any rate the implementation is consistent with regularity. Owning messages are implemented in terms of arrays; copying an array copies all the elements. So that's what we're using for quality.

elbeno: Resolved offline.

mjcaisse-intel (approved):

PR #591, opened 2024-08-06
encourage the compiler to inline flow steps

lukevalenty:


include/flow/impl.hpp
```diff
@@ -81,16 +81,16 @@ class impl : public interface {
...
         if constexpr (loggingEnabled) {
             CIB_TRACE("flow.start({})", name);
         }
 
-        for (auto const func : functionPtrs) {
-            func();
-        }
+        [this]<std::size_t... Is>(std::index_sequence<Is...>){
+            (functionPtrs[Is](), ...);
+        }(std::make_index_sequence<capacity>{});
```
elbeno, line 93: **suggestion**: Make the unrolling a policy?

elbeno (approved):

include/flow/impl.hpp
lukevalenty: I think we _always_ want this. This should work for both clang and gcc. If we find a need later for another compiler or version then I think we could add it then.

PR #602, opened 2024-08-25
enable byte indexing for msg::located at

mjcaisse:


include/msg/field.hpp
```diff
@@ -323,6 +332,34 @@ template <> struct at<dword_index_t, msb_t, lsb_t> {
     }
 };
 
+template <> struct at<byte_index_t, msb_t, lsb_t> {
+    byte_index_t index_{};
+    msb_t msb_{};
+    lsb_t lsb_{};
+
+    [[nodiscard]] constexpr auto
+    index() const -> std::underlying_type_t<byte_index_t> {
```
elbeno, line 341: **suggestion**: If we effectively return `index_/4` here there is no need to change `bits_locator_t` and no need to introduce `indexed_bit_size`. This change can be much simpler?

include/msg/field.hpp
```diff
@@ -277,12 +279,16 @@ template <bits_locator... BLs> struct field_locator_t {
...
 enum struct lsb_t : std::uint32_t {};
 
 inline namespace literals {
 // NOLINTNEXTLINE(google-runtime-int)
+CONSTEVAL auto operator""_bi(unsigned long long int v) -> byte_index_t {
+    return static_cast<byte_index_t>(v);
+}
 CONSTEVAL auto operator""_dw(unsigned long long int v) -> dword_index_t {
     return static_cast<dword_index_t>(v);
 }
```
elbeno, line 294: **suggestion**: Let's find a consistent UDL format here. `_dword` and `_byte` for example. Or possibly `_dwi` and `_bi`. But not one thing as "index" and the other as "datatype" if you see what I'm saying.

(We will probably have to leave `_dw` as is, but we could make `_dword` as well.)

include/msg/field.hpp
mjcaisse: Are you suggesting to also modify `lsb()`, `size()`, and `sort_key()` so that everything stays in 32-bit words? I was thinking of normalizing everything to bytes...  (o;

include/msg/field.hpp
mjcaisse: I prefer the "index" over the "datatype" but wasn't wild about `_bi`. You already know I hate `dword` for all things.  (o;  I'll sleep on this.

include/msg/field.hpp
mjcaisse: updated

include/msg/field.hpp
mjcaisse: I went `_bi` and `_dwi` for now.

mjcaisse:
Just realized I didn't update the docs.

include/msg/field.hpp
elbeno: I'm suggesting we decouple the interface of `bits_locator_t` from `at` - before this PR, `bits_locator_t` expects things in dwords, but it might make sense for it to work in bytes.

elbeno (approved):

PR #603, opened 2024-08-27
:sparkles: Add `can_hold` to `field`

elbeno:
A field's type can be larger than the bits used in the message. It is sometimes useful at runtime to determine whether or not a given value will fit in a field.

### DIFF
--- a/include/msg/field.hpp
+++ b/include/msg/field.hpp
@@ -274,6 +274,8 @@ template <bits_locator... BLs> struct field_locator_t {
     template <typename T> constexpr static auto extent_in() -> std::size_t {
         return std::max({std::size_t{}, BLs::template extent_in<T>()...});
     }
+
+    constexpr static auto size = (std::size_t{} + ... + BLs::size);
 };
 } // namespace detail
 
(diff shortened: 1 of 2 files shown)


include/msg/field.hpp
```diff
@@ -451,6 +453,10 @@ class field_t : public field_spec_t<Name, T, detail::field_size<Ats...>>,
         return format("{}: 0x{:x}"_sc, spec_t::name, v);
     }
 
+    constexpr static auto can_hold(value_type v) -> bool {
```
lukevalenty, line 456: question: i wonder if it's worthwhile to have an `if constexpr` shortcut that checks if the type can fit at compile-time or if the compiler will probably do this

lukevalenty (approved):

include/msg/field.hpp
bdeane-intel: It's a constexpr function; if you have a compile-time value to put in the field you can just `static_assert` the predicate.

bdeane-intel: The static invariant on fields is that the field's type can represent the number of bits in the message. The reverse is not necessarily true, on purpose. The field's type is often a perf choice where the number of bits is of course a protocol space choice.

lukevalenty: i'm talking about the case where it's still a runtime value, but the type of the runtime value will always fit within the field

bdeane-intel: If you have a compile time value, you can static_assert that it fits. If you have a runtime value, you can test. If you have a runtime value that somehow always fits, you are in the realm of contracts and either you choose to check it (at runtime) or not (accept UB on fail)?

bdeane-intel: After offline discussion - this is the scenario: https://godbolt.org/z/G7rG3oaTM
And the optimizer can see through.

PR #605, opened 2024-08-27
enable more aggressive inlining of flows

lukevalenty:


include/flow/graph_builder.hpp
```diff
@@ -149,7 +149,14 @@ struct graph_builder {
             constexpr auto built = build(v);
             static_assert(built.has_value(),
                           "Topological sort failed: cycle in flow");
-            return *built;
+
+            constexpr auto functionPtrs = built->functionPtrs;
+            constexpr auto size = functionPtrs.size();
+            constexpr auto name = built->name;
+
+            return [&]<std::size_t... Is>(std::index_sequence<Is...>) {
```
elbeno, line 157: **thought**: I don't think the lambda needs to capture anything.

test/flow/graph_builder.cpp
```diff
@@ -6,119 +6,140 @@
...
+[[maybe_unused]] constexpr auto milestone1 = flow::milestone<"milestone1">();
 
-constexpr auto a = flow::action<"a">([] { actual += "a"; });
-constexpr auto b = flow::action<"b">([] { actual += "b"; });
-constexpr auto c = flow::action<"c">([] { actual += "c"; });
-constexpr auto d = flow::action<"d">([] { actual += "d"; });
+[[maybe_unused]] constexpr auto a = flow::action<"a">([] { actual += "a"; });
+[[maybe_unused]] constexpr auto b = flow::action<"b">([] { actual += "b"; });
+[[maybe_unused]] constexpr auto c = flow::action<"c">([] { actual += "c"; });
+[[maybe_unused]] constexpr auto d = flow::action<"d">([] { actual += "d"; });
```
elbeno, line 17: **question**: Are all these `[[maybe_unused]]`s needed, or are they an artifact of getting the tests to work that can now be removed?

include/flow/graph_builder.hpp
lukevalenty: tried without the capture, it needs it for the `constexpr` variables....even though they are `constexpr` :(

test/flow/graph_builder.cpp
lukevalenty: they are an artifact...will remove

elbeno (approved):

PR #608, opened 2024-08-30
:bug: Fix handler short-circuiting

elbeno:
The handler should call all the callbacks that match a message, not just the first that matches.

This is consistent with the indexed_handler behaviour.

NOTE: The order of calling the callbacks is not defined.


lukevalenty (approved):
Nice, thanks Ben!

PR #614, opened 2024-09-11
:art: Allow messages to be handled by value

elbeno:
Messages under a certain size are fine to copy and pass in registers.


lukevalenty:
I'm not quite following how the code works...does the change in this PR pass the message by value (owning) when the message is under a certain size?

bdeane-intel:
It doesn't, it just allows the author of the callback/matcher to take the message by value. They'd make the decision to do that based on size.

lukevalenty (approved):

PR #615, opened 2024-09-11
:art: Allow indexed handler to handle messages by value

elbeno:
- remove the `T` suffix from several template arguments
- rename `BaseMsg` to `MsgBase` where appropriate: there was a mix of uses, and it isn't necessarily a message at all, but the raw storage underlying. So it's not a base message, but it is a message base.
- handle by-value callbacks in the indexed builder/handler.

### DIFF
--- a/include/msg/handler_builder.hpp
+++ b/include/msg/handler_builder.hpp
@@ -6,21 +6,20 @@
 #include <stdx/tuple_algorithms.hpp>
 
 namespace msg {
-template <typename CallbacksT, typename BaseMsgT,
-          typename... ExtraCallbackArgsT>
+template <typename Callbacks, typename MsgBase, typename... ExtraCallbackArgs>
 struct handler_builder {
-    CallbacksT callbacks;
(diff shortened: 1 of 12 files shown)


include/msg/handler_builder.hpp
```diff
@@ -6,21 +6,20 @@
 #include <stdx/tuple_algorithms.hpp>
 
 namespace msg {
-template <typename CallbacksT, typename BaseMsgT,
-          typename... ExtraCallbackArgsT>
+template <typename CallbacksT, typename MsgBase, typename... ExtraCallbackArgs>
```
mjcaisse-intel, line 9: todo: `Callbacks` instead of `CallbacksT`

mjcaisse-intel (approved):

include/msg/handler_builder.hpp
elbeno: Good catch - somehow I missed that one instance.

PR #641, opened 2024-10-30
fix: base_extent calculation in bits_locator_t::template extent_in()

samvangysegem:
The `base_extent` variable in the `extent_in()` method of the `bits_locator_t` struct now uses the size of its template parameter T instead of `std::uint32_t` for calculating the base extent .


elbeno:
I don't think this is right - the `Index` is always in dwords, hence computing the `extent` in bytes uses `sizeof(std::uint32_t)`.

PR #644, opened 2024-11-14
:sparkles: Fine-grained logging control for flows

elbeno:


include/flow/log.hpp
```diff
@@ -0,0 +1,32 @@
...
+template <typename T, typename... DummyArgs>
+    requires(sizeof...(DummyArgs) == 0)
+constexpr static auto get_log_spec() {
+    if constexpr (std::is_same_v<decltype(log_spec<T::ct_name, DummyArgs...>),
+                                 default_log_spec const>) {
+        return log_spec<"default", DummyArgs...>;
+    } else {
+        return log_spec<T::ct_name, DummyArgs...>;
+    }
+}
```
lukevalenty: question: is there a way for a user of the library to specify the default log level?

lukevalenty (approved):

include/flow/log.hpp
elbeno: Yes - shown in the "override default log level" test.

PR #646, opened 2024-11-18
:art: :boom: Apply consistent naming between callback and flow

elbeno:
Problem:
- `flow` constructs are called `flow::builder` and `flow::service`.
- `callback` constructs are called `cib::callback` and `cib::callback_meta`, which is confusing.

Solution:
- Rename the `callback` constructs to `callback::builder` and `callback::service` to match the `flow` pattern.


lukevalenty (approved):
i like it!

PR #647, opened 2024-11-18
:art: Collapse flow step run/log functions

elbeno:
Problem:
- Now that built flow steps have been refactored to have access to the right compile-time information, there's no need to have separate `run` and `log` functions.

Solution:
- Collapse built flow steps so that one function logs the step and runs it.

### DIFF
--- a/include/flow/impl.hpp
+++ b/include/flow/impl.hpp
@@ -2,109 +2,48 @@
 
 #include <flow/common.hpp>
 #include <flow/log.hpp>
-#include <flow/step.hpp>
 #include <log/log.hpp>
 
 #include <stdx/ct_string.hpp>
-#include <stdx/cx_vector.hpp>
 #include <stdx/span.hpp>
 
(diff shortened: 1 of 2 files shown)


include/flow/impl.hpp
```diff
@@ -2,109 +2,48 @@
...
 #include <cstddef>
 #include <iterator>
-#include <type_traits>
 
 namespace flow {
 namespace detail {
-template <typename CTNode> constexpr auto run_func() -> void {
+template <stdx::ct_string FlowName, typename CTNode>
+constexpr auto run_func() -> void {
     if (CTNode::condition) {
```
lukevalenty, line 19: issue: `CTNode::condition` needs to be invoked, right?

lukevalenty: omg, I forgot my own code that I wrote! XD

lukevalenty (approved):

PR #658, opened 2024-12-09
Split off a few libraries

elbeno:
In general, CIB does not have a well-defined dependency DAG between its sublibraries; but three are able to be split off quite easily: `sc`, `match` and `interrupt`.


lukevalenty (approved):
awesome! i was glad to see the `cib` library still contains everything.

PR #663, opened 2024-12-28
:art: Clean up pseudo pext benchmarks

elbeno:
See #561


lukevalenty (approved):
Thanks for splitting it into the different commits! 😎

PR #672, opened 2025-01-03
:bug: Fix handling optional stable catalog

elbeno:
Problem:
- When there are no stable strings, one cannot get them from a dictionary.

Solution:
- Provide default empty lists.

### DIFF
--- a/tools/gen_str_catalog.py
+++ b/tools/gen_str_catalog.py
@@ -1,11 +1,11 @@
 #!/usr/bin/env python3
 
 import argparse
-from functools import partial
 import itertools
 import json
 import re
 import xml.etree.ElementTree as et
+from functools import partial
 
 
 def find_arg_split_pos(s: str, start: int) -> int:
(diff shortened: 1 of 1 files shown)


tools/gen_str_catalog.py
```diff
@@ -215,7 +215,7 @@ def write_json(messages, modules, extra_inputs: list[str], filename: str, stable
 
 
 def read_stable(stable_filenames: list[str]):
-    stable_catalog = dict()
+    stable_catalog: dict = dict(messages=[], modules=[])
```
lukevalenty, line 218: question: why not use additional type hint information like `dict[str, list[str]]`?

tools/gen_str_catalog.py
```diff
@@ -397,10 +397,12 @@ def main():
     args = parse_cmdline()
 
     stable_catalog = read_stable(args.stable_json)
+    stable_messages = stable_catalog.get("messages", [])
```
lukevalenty, line 400: suggestion: maybe `collections.defaultdict` would be nicer

lukevalenty (approved):

tools/gen_str_catalog.py
elbeno: I made it a `dict[str, list]` - I'm loath to overly constrain the type of the values inside the lists. They're really just JSON. If we had concepts, maybe we could constrain them to contain the keys we're expecting...

tools/gen_str_catalog.py
elbeno: Actually this isn't needed because of the defaulting above. So I removed this.

PR #673, opened 2025-01-08
:sparkles: Add `relaxed_message`

### LINKED ISSUE #664: ✨ `msg::relaxed_message`: Add a `msg` type that lets one or more field locations be allocated automatically
Sometimes we don't know exactly what a new message format will look like. We may have a pretty good idea of the fields and their types, but not necessarily how they will be packed together. It would be helpful to have a `msg::relaxed_message` type that lets us define such a message. 

As the exact layout of the message is defined, more and more field locations can be specified. Once they are all specified, then an author may choose to change it to a `msg::message` that requires all field locations to be explicitly defined.

```c++
    using my_msg_defn =
        msg::relaxed_message<
            "my_msg",
            msg::field<"src", std::uint8_t>::located<at{dw_{0}, msb_{ 7}, lsb_{0}}>,
            msg::field<"dst", std::uint8_t>::located<at{dw_{0}, msb_{15}, lsb_{8}}>,
            msg::field<"addr", std::uint32_t>,
            msg::field<"data", std::uint32_t>>;
```

Additionally, some messages don't need an explicit layout and it's fine letting the library pack them for us.

elbeno:
Problem:
- Developing message formats is onerous; we'd like the compiler to take care of packing messages until we know exactly where to put fields.

Solution:
- Add `relaxed_message` as in #664

Note:
- All auto-located fields are put after the fixed spec fields.
- Auto-located fields are sorted from largest to smallest to ease alignment.

Closes #664

### DIFF
--- a/include/msg/field.hpp
+++ b/include/msg/field.hpp
@@ -532,5 +532,8 @@ struct field {
     using located = detail::field_t<
         decltype(stdx::ct_string_to_type<Name, sc::string_constant>()), T,
         Default, M, Ats...>;
+
+    constexpr static auto bitsize = sizeof(T) * CHAR_BIT;
+    using default_located = located<at{msb_t{bitsize - 1}, lsb_t{}}>;
 };
 } // namespace msg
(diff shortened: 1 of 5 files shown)


mjcaisse-intel (approved):
Looks like a nice addition!

PR #679, opened 2025-01-12
:sparkles: Introduce log environment

elbeno:
Problem:
- Logging is parameterized on many axes. Most logging systems include the level.
  Some also include a module or source. Some include other parameters.
- Logging machinery should be agnostic to the parameterizations of each logging
  backend. Adding extra parameters to logging macros is not scalable.

Solution:
- Expand the current logging module to the idea of a logging environment. An
  environment allows compile-time scope-based arguments to be passed through the
  logging machinery; logging backends can extract whichever values they can use
  from the current environment.


lukevalenty (approved):
🔥

PR #688, opened 2025-02-11
:sparkles: Allow overriding the mipi message type

elbeno:
Problem:
- Mipi short messages do not contain anything except the string id: no module id, no severity, etc. A client may want this information, so may not want to automatically use the short message.

Solution:
- Allow the logging environment to supply a message builder for mipi.


mjcaisse-intel (approved):
nice!

PR #691, opened 2025-02-18
:art: Remove inlining attributes from logging functions

elbeno:
Problem:
- The interplay of inlining and outlining is too complex. CE experiments show that the compiler does the right thing without it all.

Solution:
- Remove it all.


mjcaisse-intel (approved):
o.O   looking forward to seeing how this performs in our projects

PR #692, opened 2025-02-19
:art: Allow flow steps to override log environment

elbeno:
Problem:
- Flow steps allow specialization of a log spec that can contain a custom level and flavor, but not the general mechanism available to log environments.

Solution:
- Change a flow log spec to a log environment.


mjcaisse-intel (approved):
sweet!

PR #696, opened 2025-02-25
:sparkles: Add enum information to string catalog json output

elbeno:


elbeno:
The resulting section in the JSON looks like this:

```javascript
"enums": {
    "ns::E": {
        "42": "value"
    }
}
```

@siauthun Note that the enumeration is a dictionary, not a list (different from what we originally talked about yesterday). In general we can't assume that enumeration values start at zero and increment uniformly.

lukevalenty (approved):

PR #698, opened 2025-02-26
:bug: Allow 64-bit values in MIPI arg encoding

elbeno:
Problem:
- All runtime arguments to logs are assumed by the mipi encoder to be `std::uint32_t`.
- In reality, the mipi spec supports 64-bit values as well. (See MIPI-SyS-T spec table 16.) In particular, logging a time value (from a `std::chrono::duration`) needs to account for 64 bits.

Solution:
- Pack integral values properly according to the spec. Anything 32 bits or less is packed as 32 bits. If it's 64-bits, it's packed that way.
- The XML output strings use the appropriate (printf-like) specifiers.
- The JSON output retains the C++ types, as does the generated C++.

Notes:
- Table 16 allows 64-bit integral values even when the catalog message subtype is *_P32.
- Floating point and pointer values are supported by MIPI but not by this code. Maybe in future, if there is a use case.

### DIFF
--- a/test/log/mipi_encoder.cpp
+++ b/test/log/mipi_encoder.cpp
@@ -66,6 +66,7 @@ struct test_conc_policy {
 int num_log_args_calls{};
 
 constexpr auto check = [](auto value, auto expected) {
+    static_assert(std::is_same_v<decltype(value), decltype(expected)>);
     CHECK(value == expected);
 };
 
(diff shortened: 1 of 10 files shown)


test/log/mipi_encoder.cpp
```diff
@@ -66,6 +66,7 @@ struct test_conc_policy {
 int num_log_args_calls{};
 
 constexpr auto check = [](auto value, auto expected) {
+    // static_assert(std::is_same_v<decltype(value), decltype(expected)>);
```
lukevalenty, line 69: todo: remove or replace comment

tools/gen_str_catalog.py
```diff
@@ -267,9 +271,9 @@ def serialize_messages(short_node: et.Element, catalog_node: et.Element, message
                 "EnumLookup", f'{msg["enum_lookup"][0]}:{msg["enum_lookup"][1]}'
             )
 
-        fmt_string = re.sub(r"{}", r"%d", msg["msg"])
-        if msg["arg_count"] != 0:
-            fmt_string = re.sub(r"{:(.*?)}", r"%\1", fmt_string)
+        fmt_string = msg["msg"]
+        for arg in msg["arg_types"]:
+            fmt_string = re.sub(r"{.*?}", arg_printf_spec(arg), fmt_string, count = 1)
```
lukevalenty, line 276: issue: we need to keep the format specifiers

lukevalenty (approved):

PR #702, opened 2025-03-06
Add message environments

### LINKED ISSUE #684: 🎇 Compile-time attributes for messages
It would be helpful to have compile-time attributes for messages. For example, some interfaces send messages in posted or non-posted manner. Some messages may _require_ to be sent in one way or the other. 

If message definitions contained compile-time fields, then this information could be coded and checked that way.

elbeno:
Add (optional) compile-time attributes for messages using `stdx::env`.
Use a message environment to control logging for a callback match.

Closes #684


lukevalenty (approved):
🔥

PR #707, opened 2025-03-13
:white_check_mark: Improve log FATAL tests

elbeno:


test/log/log.cpp
```diff
@@ -31,34 +42,57 @@ inline auto logging::config<> =
 template <> inline auto stdx::panic_handler<> = injected_handler{};
 
 TEST_CASE("CIB_FATAL logs the string", "[log]") {
+    expected_why = {};
```
mjcaisse-intel, line 45: thought: these three activities happen each time ... does it make sense to just make a reset function?

mjcaisse-intel (approved):

PR #714, opened 2025-03-26
:art: Unify message callbacks between indexed and normal handlers

elbeno:
Problem:
- The indexed handler callback duplicates code from the regular callback `handle` function, just for the purpose of adding a little to the logged message.

Solution:
- Allow an extra string to be passed to the regular callback `handle`.


mjcaisse-intel (approved):
nice simplification

PR #716, opened 2025-03-27
:bug: Field matcher logs shouldn't coerce args to `std::uint32_t`

elbeno:
Problem:
- When field matchers log their match descriptions, they coerce the field types to std::uint32_t. Logging can deal with other types now.

Solution:
- Preserve the field types. Only log hex values for integral types.


lukevalenty (approved):
Awesome 👏🏻

PR #722, opened 2025-04-16
:sparkles: Add typo detection/fixing for string generation

elbeno:
Problem:
- Stable strings can be fixed not just to be consistent from one build to the next, but because hardware expects certain string IDs or ranges. Accidentally having a string in code with a typo could result in another string ID being generated, with unfortunate results.

Solution:
- Allow typos to be detected and fixed.

### DIFF
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -126,13 +126,13 @@ jobs:
       - name: Install build tools
         run: |
           ${{ matrix.install }}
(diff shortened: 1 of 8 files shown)


elbeno:
You're saying you want an additional `stable_only` policy that forces every string to be in `stable_strings.json`?

elbeno:
This PR is abandoned. Typo detection doesn't really solve any problem we have in the right way.
The useful CI part is captured in #726

PR #725, opened 2025-05-07
:bug: Make `field_name` assignment operation `const`

elbeno:
Problem:
- `field_name` is an empty type, and `field_name` assignment is an overloaded operator that returns a different type. Without making it `const`-qualified it cannot be called on template arguments.

Solution:
- `const`-qualify `field_name::operator=`.


elbeno:
@jvillavc this enables use of field_name values as template arguments.

lukevalenty (approved):

PR #727, opened 2025-05-14
:sparkles: Add fixed string IDs to logs

elbeno:
Problem:
- Some hardware expects certain string IDs. When logging, it would be useful to be able to fix string IDs in code and have the tooling understand that.

Solution:
- Add the ability to specify string IDs in the logging environment.

### DIFF
--- a/include/log/catalog/encoder.hpp
+++ b/include/log/catalog/encoder.hpp
@@ -4,6 +4,7 @@
 #include <log/catalog/catalog.hpp>
 #include <log/log.hpp>
 #include <log/module.hpp>
+#include <log/string_id.hpp>
 
 #include <stdx/ct_string.hpp>
 #include <stdx/span.hpp>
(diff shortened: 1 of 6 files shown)


include/log/catalog/encoder.hpp
```diff
@@ -85,7 +88,8 @@ template <typename Writer> struct log_handler {
             auto builder = get_builder(Env{});
             constexpr auto L = stdx::to_underlying(get_level(Env{}));
             using Message = typename decltype(builder)::template convert_args<
-                detail::to_message_t<decltype(fr.str)>::template fn,
+                detail::to_message_t<decltype(fr.str), logging::get_string_id(
+                                                           Env{})>::template fn,
```
bdeane-intel, line 92: **note**: It would be possible to lift the ID straight from the environment here, if it exists. But doing the round-trip through the tooling with the `catalog<>()` specialization + LTO is also fine, and makes for a single path for all strings.

bdeane-intel:
The ad-hoc data structures in `gen_str_catalog.py` have been... unwieldy... for a while. So I opened #728.

lukevalenty (approved):

PR #738, opened 2025-06-15
:books: Clarify field documentation

elbeno:
Problem:
- The docs for `field` mention the storage in 32-bit words, but this isn't really applicable to a field; it's a message thing (and is mentioned later in that context).
- Field locators may omit dword indices, but this is not documented.
- A couple of important things about field size are not documented.

Solution:
- Remove the reference to storage in the field section.
- Add a note that the max field size is 64 bits.
- Document the alternative `located` specialization.
- Note the compilation error when storage size exceeds type representation.

### DIFF
--- a/docs/message.adoc
+++ b/docs/message.adoc
@@ -8,8 +8,7 @@ Everything in the msg library is in the `msg` namespace.
 === Fields
 
 A `field` represents a value, specified in bits, inside a unit of addressable
-storage. Currently in the _cib_ library, storage is assumed to be in units of
-`std::uint32_t`. A field is a view type: it defines how to access storage, but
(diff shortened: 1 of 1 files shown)


docs/message.adoc
```diff
@@ -155,7 +181,7 @@ using msg_defn = extend<
 ----
 
 The combined definition incorporates all the fields of the messages. And as
-shown, the combination might typically be `extend`ed with a constraint on the
+shown, the combination might typically be `extend`​ed with a constraint on the
```
elbeno, line 184: **note**: The diff here is the addition of a zero-width space immediately after \`extend\` to make the asciidoctor rendering correct.

lukevalenty (approved):

PR #749, opened 2025-08-05
:sparkles: Add target for string collateral

elbeno:
Problem:
- The outputs produced by `gen_str_catalog` are built by a custom command but do not have an associated target, which means other targets can't depend on them.

Solution:
- Add an optional target for the `.cpp`, `.xml` and `.json` files produced by `gen_str_datalog`.


mjcaisse-intel (approved):
Thanks!

PR #771, opened 2025-09-27
:books: Separate mermaid diagrams

elbeno:
Problem:
- Mermaid diagrams are inlined in documentation, which means Github does not know how to render them.

Solution:
- Separate mermaid docs into their own files. Github knows how to render `.mmd` files.


elbeno:
TIL Github can render .mmd files: https://github.com/elbeno/compile-time-init-build/blob/separate-diagrams/docs/library_deps.mmd

PR #779, opened 2025-10-06
:memo: Add logging examples

elbeno:
Problem:
- There are no examples for various use cases of logging.

Solution:
- Add examples.


mjcaisse-intel (approved):
:praise: Thank you for these examples!

PR #796, opened 2025-11-04
cmake: use ${PROJECT_NAME}_IS_TOP_LEVEL instead of PROJECT_IS_TOP_LEVEL

benedekkupper:
I use FetchContent_Declare() and FetchContent_MakeAvailable() to pull in the cib dependency to an ESP-IDF project (among others). Strangely, at least with cmake version 3.28.3, PROJECT_IS_TOP_LEVEL evaluates to true in such scenarios. The outcome of the mixup is a completely broken build. Using ${PROJECT_NAME}_IS_TOP_LEVEL solves the problem, as it produces the expected behavior.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -9,7 +9,7 @@ project(
     HOMEPAGE_URL "https://github.com/intel/compile-time-init-build")
 
 include(cmake/get_cpm.cmake)
-if(PROJECT_IS_TOP_LEVEL)
+if(${PROJECT_NAME}_IS_TOP_LEVEL)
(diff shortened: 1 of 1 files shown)


elbeno:
I think the issue is in https://github.com/espressif/esp-idf/blob/master/tools/cmake/project.cmake#L602 - this attempts to restore the `project` function and propagates a bunch of variables set by it, but doesn't seem to set anything to do with `PROJECT_IS_TOP_LEVEL`.

I think your fix is working by accident. It looks to me like  `${PROJECT_NAME}_IS_TOP_LEVEL` is not actually false here: I think it's probably unbound.

PR #800, opened 2025-11-11
:sparkles: Generate flow diagram from code non-invasively

elbeno:
Problem:
- In order to debug a flow, we can use an alternative renderer, but it requires changing code.

Solution:
- Add `generate_flow_graph` which injects the debug renderer for a particular named flow, to produce a compile-time error containing the appropriate part of the flow. Mermaid can then be used to render that output into an svg file.

### DIFF
--- a/cmake/debug_flow.cmake
+++ b/cmake/debug_flow.cmake
@@ -1,0 +1,108 @@
+if(COMMAND generate_flow_graph)
+    return()
+endif()
+
+function(generate_flow_graph)
+    cmake_parse_arguments(FG "" "TARGET" "" ${ARGN})
+    message(
+        STATUS
+            "generate_flow_graph(${FG_TARGET}) is disabled because mmdc was not found."
+    )
+endfunction()
+
+find_program(MMDC_PROGRAM "mmdc")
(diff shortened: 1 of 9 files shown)


cmake/debug_flow.cmake
```diff
@@ -0,0 +1,108 @@
...
+    set(debug_target "${target_stem}.bin")
+    make_flow_debug_target(${FG_TARGET} ${debug_target})
+    target_link_libraries(${debug_target} PRIVATE ${injection_lib})
+
+    # generate the graph source
+    set(graph_stem "${target_stem}.graph")
+    set(graph_source "${graph_stem}.mmd")
+    set(graph "${graph_stem}.svg")
+    set(awk_gcc_cmd "awk '{gsub(\"\\\\\\\\012\",\"\\n\")}\;1'")
+    set(awk_clang_cmd "awk '{gsub(\"\\\\\\\\n\",\"\\n\")}\;1'")
```
elbeno: **note**: Munging the compiler output is awkward: the good news is that from clang-20, no munging is necessary. These `awk` commands exist to deal with gcc and clang pre-20.

PR #829, opened 2026-01-05
:children_crossing: Add concepts to nexus

elbeno:
Problem:
- Nexus services are not very well constrained/described in code: it is not obvious how to achieve the correct alignment of builder, interface, etc.

Solution:
- Add concepts that tie together the builder, interface and builder_meta ideas.

### DIFF
--- a/include/msg/indexed_handler.hpp
+++ b/include/msg/indexed_handler.hpp
@@ -2,17 +2,27 @@
 
 #include <msg/detail/indexed_handler_common.hpp>
 
+#include <stdx/bitset.hpp>
 #include <stdx/compiler.hpp>
 
-namespace msg {
+#include <cstdint>
 
+namespace msg {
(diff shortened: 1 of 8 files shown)


include/msg/indexed_handler.hpp
```diff
@@ -2,17 +2,27 @@
...
+#include <cstdint>
 
+namespace msg {
 template <typename... Indices> struct indices : Indices... {
     CONSTEVAL explicit indices(Indices... index_args)
         : Indices{index_args}... {}
 
     constexpr auto operator()(auto const &data) const {
-        return (this->Indices::operator()(data) & ...);
+        return (... & this->Indices::operator()(data));
```
lukevalenty: question: why this change?

elbeno: Tried something out. Ended up not going with it, but left this as a left fold for preference.

PR #879, opened 2026-04-08
Fix link to User Guide in examples README.md

adavidzh:
Existing link is broken.
This should be better than `../../../` and hopefully still work.


elbeno (approved):

adavidzh:
Also, kudos on all the developments supporting declarative C++ programming!

PR #908, opened 2026-06-01
⚡ perfrom flow graph dsl calculations in type space

mjcaisse-intel:
Problem:
- Some projects that heavily use flow can take considerable time to compile.
- `get_edges_t` CPO is effectively cubic growth in the front-end: O(N³)

Solution:
- Don't recursively `tuple_cat` at each of the expression (>>) levels of the graph.

### DIFF
--- a/include/flow/dsl/flatten.hpp
+++ b/include/flow/dsl/flatten.hpp
@@ -1,0 +1,51 @@
+#pragma once
+
+#include <flow/dsl/subgraph_identity.hpp>
+#include <flow/dsl/walk.hpp>
+
+#include <stdx/tuple.hpp>
+
+#include <boost/mp11.hpp>
+
+#include <type_traits>
+
+namespace flow::dsl::detail {
+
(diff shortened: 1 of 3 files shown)


include/flow/dsl/flatten.hpp
```diff
@@ -0,0 +1,51 @@
...
+    template <typename Pair>
+    using fn = edge<boost::mp11::mp_first<Pair>, boost::mp11::mp_second<Pair>,
+                    Cond>;
+};
+
+template <typename List> struct to_tuple;
+template <typename... Ts> struct to_tuple<boost::mp11::mp_list<Ts...>> {
+    constexpr static auto value = stdx::tuple<Ts...>{};
+};
+template <typename List> constexpr auto to_tuple_v = to_tuple<List>::value;
```
elbeno, line 49: **suggestion**: Use `boost::mp11::mp_apply<stdx::tuple, List>`

PR #918, opened 2026-06-12
⚡ perform flow graph dsl calculations in type space

mjcaisse-intel:
Problem:
- Some projects that heavily use flow can take considerable time to compile
- get_edges_t CPO is effectively cubic growth in the front-end: O(N³)

Solution:
- Don't recursively tuple_cat at each of the expression (>>) levels of the graph
- Replace CPO mechanism with simpler calls

### DIFF
--- a/include/flow/dsl/par.hpp
+++ b/include/flow/dsl/par.hpp
@@ -3,7 +3,8 @@
 #include <flow/dsl/subgraph_identity.hpp>
 #include <flow/dsl/walk.hpp>
 
-#include <stdx/tuple_algorithms.hpp>
+#include <boost/mp11/algorithm.hpp>
+#include <boost/mp11/list.hpp>
 
 namespace flow::dsl {
 template <subgraph Lhs, subgraph Rhs,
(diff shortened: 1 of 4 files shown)


include/flow/dsl/flatten.hpp
```diff
@@ -0,0 +1,130 @@
...
+    }
+}
+
+template <typename Leaves, std::size_t... Is>
+constexpr auto leaf_star_list(std::index_sequence<Is...>)
+    -> boost::mp11::mp_list<star_t<stdx::tuple_element_t<Is, Leaves>>...>;
+
+template <typename Leaves>
+using leaf_star_list_t = decltype(leaf_star_list<Leaves>(
+    std::make_index_sequence<stdx::tuple_size_v<Leaves>>{}));
```
elbeno: **suggestion**: I think this is the same as:
`boost::mp11::mp_apply<boost::mp11::mp_list, boost::mp11::mp_transform<star_t, Leaves>>`?

include/flow/dsl/flatten.hpp
elbeno, line 1: **question**: Why the name `flatten.hpp`?

include/flow/dsl/flatten.hpp
```diff
@@ -0,0 +1,130 @@
...
+template <typename N, typename Leaves>
+[[nodiscard]] constexpr auto pick_node(Leaves const &leaves) {
+    using leaves_t = stdx::remove_cvref_t<Leaves>;
+    constexpr auto idx =
+        boost::mp11::mp_find<leaf_star_list_t<leaves_t>, N>::value;
+    if constexpr (idx < stdx::tuple_size_v<leaves_t>) {
+        return *stdx::get<idx>(leaves);
+    } else {
+        return N{};
+    }
```
elbeno: **question**:  Given that these tuples are of empty types, ought we to use mp11 here (mp_size, mp_at) instead of tuple (tuple_size, get)?

include/flow/dsl/flatten.hpp
```diff
@@ -0,0 +1,130 @@
+#pragma once
+
+#include <flow/dsl/subgraph_identity.hpp>
+#include <flow/dsl/walk.hpp>
+
+#include <stdx/tuple.hpp>
+#include <stdx/tuple_algorithms.hpp>
```
elbeno: **note**: This is (probably?) a superfluous include.

include/flow/dsl/flatten.hpp
```diff
@@ -0,0 +1,130 @@
...
+
+#include <boost/mp11.hpp>
```
elbeno: **suggestion**: Avoid the omnibus include.

include/flow/dsl/par.hpp
```diff
@@ -1,10 +1,13 @@
 #pragma once
 
+#include <flow/dsl/flatten.hpp>
 #include <flow/dsl/subgraph_identity.hpp>
 #include <flow/dsl/walk.hpp>
 
 #include <stdx/tuple_algorithms.hpp>
 
+#include <boost/mp11.hpp>
```
elbeno: **suggestion**: Avoid the omnibus include.

include/flow/dsl/par.hpp
```diff
@@ -1,10 +1,13 @@
 #pragma once
 
+#include <flow/dsl/flatten.hpp>
 #include <flow/dsl/subgraph_identity.hpp>
 #include <flow/dsl/walk.hpp>
 
 #include <stdx/tuple_algorithms.hpp>
```
elbeno: **suggestion**: Remove this include.

include/flow/dsl/seq.hpp
```diff
@@ -1,11 +1,15 @@
 #pragma once
 
+#include <flow/dsl/flatten.hpp>
 #include <flow/dsl/subgraph_identity.hpp>
 #include <flow/dsl/walk.hpp>
-#include <nexus/detail/runtime_conditional.hpp>
 
 #include <stdx/tuple_algorithms.hpp>
```
elbeno: **suggestion**: Remove this include.

include/flow/dsl/seq.hpp
```diff
@@ -1,11 +1,15 @@
...
 
+#include <boost/mp11.hpp>
```
elbeno: **suggestion**: Avoid the omnibus include.

include/flow/dsl/seq.hpp
```diff
@@ -15,54 +19,53 @@ struct seq {
...
+struct nodes_of<seq<L, R, subgraph_identity::REFERENCE, C>> {
+    using type = boost::mp11::mp_append<nodes_of_t<L>, nodes_of_t<R>>;
 };
 
-template <subgraph Lhs, subgraph Rhs> seq(Lhs, Rhs) -> seq<Lhs, Rhs>;
+template <typename L, typename R, subgraph_identity I, typename C>
+struct edges_of<seq<L, R, I, C>> {
+    using pairs = boost::mp11::mp_product<boost::mp11::mp_list, finals_of_t<L>,
+                                          initials_of_t<R>>;
+    using cross = boost::mp11::mp_transform_q<make_edge_q<C>, pairs>;
```
elbeno: **suggestion**: How about `mp_product_q<make_edge_q<C>, finals_of_t<L>, initials_of_t<R>>`?

include/flow/dsl/flatten.hpp
```diff
@@ -0,0 +1,130 @@
...
+};
+
+template <typename List>
+constexpr auto to_tuple_v = boost::mp11::mp_apply<stdx::tuple, List>{};
+
+// instantiate all the leaf values and preserve the runtime values
+template <typename N>
+[[nodiscard]] constexpr auto collect_node_values(N const &n) {
+    // builders just hold lists of graphs... recurse
+    if constexpr (requires { typename N::is_builder; }) {
```
elbeno: **thought**: Think I would prefer it if we were telling here rather than asking.

include/flow/dsl/flatten.hpp
mjcaisse-intel: needed for `tuple_cat` in the `collect_node_values` methods

include/flow/dsl/flatten.hpp
mjcaisse-intel: )o:   This is really a residual when working through how to flatten things in type space. I shouldn't be adding another file, just editing the existing walk. Good catch.

include/flow/dsl/seq.hpp
mjcaisse-intel:
```c++
template<class Q, class... L> using mp_product_q = mp_product<Q::template fn, L...>;
```
`pairs` is a `mp_list` of the product pairs. If I understand correctly, using `mp_transform_q` would require that the quoted meta function take a list of products and then transform the list with `make_edge_q<C>`. I don't think this is a direct replacement.

elbeno: You would have to adjust `make_edge_q` so that its `fn` was a binary function.

elbeno: IOW `pairs` is an `mp_list` because `finals_of_t<L>` is an `mp_list`, not because we are passing `mp_list` as the function. If we pass `make_edge_q` as the (quoted) function it will make a list of edges directly, instead of a list of pairs that we transform into a list of edges.

mjcaisse-intel: Thank you for the mp11 education Ben!

elbeno (approved):

include/flow/dsl/flatten.hpp
elbeno: Resolution: they aren't empty types.

PR #946, opened 2026-07-17
⚡ perform flow graph dsl calculations in type space

mjcaisse-intel:
Problem:

- Some projects that heavily use flow can take considerable time to compile
- get_edges_t CPO is effectively cubic growth in the front-end: O(N³)

Solution:

- Don't recursively `tuple_cat` at each of the expression (>>) levels of the graph
- Replace CPO mechanism with simpler calls

### DIFF
--- a/include/flow/dsl/walk.hpp
+++ b/include/flow/dsl/walk.hpp
@@ -5,8 +5,14 @@
 #include <stdx/concepts.hpp>
 #include <stdx/ct_string.hpp>
 #include <stdx/tuple.hpp>
+#include <stdx/tuple_algorithms.hpp>
 #include <stdx/type_traits.hpp>
 
+#include <boost/mp11/algorithm.hpp>
+#include <boost/mp11/list.hpp>
+
+#include <cstddef>
+#include <type_traits>
 #include <utility>
 
 namespace flow::dsl {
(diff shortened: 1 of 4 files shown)


include/flow/dsl/walk.hpp
```diff
@@ -19,99 +25,129 @@ template <typename Source, typename Dest, typename Cond> struct edge {
...
-        noexcept(noexcept(tag_invoke(std::declval<get_edges_t>(),
-                                     std::forward<Ts>(ts)...)))
-            -> decltype(tag_invoke(*this, std::forward<Ts>(ts)...)) {
-        return tag_invoke(*this, std::forward<Ts>(ts)...);
+// Find the node in the existing leaf instantiations
+template <typename N, typename Leaves>
+[[nodiscard]] constexpr auto pick_node(Leaves const &leaves) {
+    using leaves_t = stdx::remove_cvref_t<Leaves>;
+    using leaf_instance_list =
+        boost::mp11::mp_apply<boost::mp11::mp_list,
```
elbeno: **suggestion**: I don't think the `mp_apply` is necessary here?

elbeno (approved):

include/flow/dsl/walk.hpp
mjcaisse-intel: simplified

PR #950, opened 2026-07-24
callback: forward rvalue args

benedekkupper:
Preserve value category in callback::builder::run so services taking T&& receive rvalues instead of lvalues.
Add a regression test for callback::service<MoveOnlyPayload&&>.

### DIFF
--- a/include/nexus/callback.hpp
+++ b/include/nexus/callback.hpp
@@ -35,7 +35,9 @@ template <typename Funcs = stdx::tuple<>, typename... ArgTypes> struct builder {
     constexpr static void run(ArgTypes... args) {
         constexpr auto b = BuilderValue::value;
         [&]<std::size_t... Is>(std::index_sequence<Is...>) {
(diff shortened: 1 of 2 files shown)


elbeno:
The reason it didn't do this before is that it's inside a fold expression. If there are multiple calllbacks you end up with a use-after-move.

benedekkupper:
> The reason it didn't do this before is that it's inside a fold expression. If there are multiple calllbacks you end up with a use-after-move.

This is a valid concern and one that I pondered too, but without a better alternative in mind I went with it anyway (just adding a comment at the callback type definition). For correctness though it should either allow rvalue forwarding, or explicitly forbid it by a `requires` clause.

elbeno:
Thanks for the contribution! Based on the discussion here I fixed this problem in a slightly more robust way in #953.

benedekkupper:
I love the solution, thanks!

PR #961, opened 2026-09-03
Correct expected output for unavailable clang module case

jefftrull:
When the clang module is not available, `catalog_output_test` will fail. It looks like simple copy-paste errors. Probably not covered by CI where (I assume) the environment always has Clang.


elbeno (approved):

elbeno:
Thanks @jefftrull!

PR #965, opened 2026-09-04
Require Python 3.12 to support backslashes in f-strings

jefftrull:
gen_str_catalog.py uses a newer feature from PEP 701 starting in commit 1e08177 that permits backslashes in f-strings. This feature is first supported in Python 3.12 from 2023. Explicitly calling out the version requirement may save users with older installations some time.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -16,7 +16,7 @@ else()
 endif()
 
 find_package(
-    Python3
+    Python3 3.12
     COMPONENTS Interpreter
     REQUIRED)
 


elbeno:
The version is already required (https://github.com/intel/compile-time-init-build/blob/main/pyproject.toml#L8).
The recommended (not documented or recommended at all, sorry, I'll fix that, but it's what CI does) way to use CIB is with a python venv controlled by `uv`.

I am not sure we want to bake an exact requirement like this into cmake as well; this should be a second-order effect?

jefftrull:
Thanks for the explanation. We didn't notice that (and have created our own CMake-based build system).

jefftrull:
I should probably elaborate on why Intel's build system setup is bypassed for our situation:

1. It's new, we don't understand it, and we're in a hurry
2. Builds generally happen on servers without Internet access, for IP-related reasons

I think both things are fairly common in the embedded space. If there is a methodology that can cleanly adapt the second constraint to the CIB build system I would like to hear more about it.

elbeno:
Re 1, I'm not sure what to do other than use as small a number of generally accepted tools as possible. Hence `cmake` and `uv`...

Re 2, the normal thing would be to vendor in your versions and/or provide them locally cached? I know this works easily with `CPM`, I am less experienced with `uv`. If there is something we're doing that prevents it, we can address that.

PR #968, opened 2026-09-06
Work around clang python binding enum_value bug

### LINKED ISSUE #964: Log: Decoding of bools is inconsistent
`log_decode.py` renders runtime bool values in some odd ways:

- plain bools are printed as `0` for false and `-1` for true, even when the underlying value in the binary file is `1` for true. This is confusing for users.
- enum bools are printed using their enum value when the underlying value is false but with `static_cast<Enum>(-1)` for true.

In other words:

| Type | Value | Rendering |
| --- | --- | --- |
| plain bool | false | 0 |
| plain bool | true | -1 |
| enum bool | false | False |
| enum bool | true | static_cast\<Enum\>(-1) |

My preference would be to use `true` and `false` for plain runtime bools (matching the treatment of the constexpr values), and the appropriate enum values for both true and false in that case.

I can put together a testcase if needed but maybe the description will suffice.

jefftrull:
At the moment, bool enums with an underlying value of true are given a value of -1 rather than +1 by the clang Python bindings. See https://github.com/llvm/llvm-project/issues/221326 for details.

This workaround intercepts -1 values returned by enum_value when the underlying type is bool, and replaces them with +1.

If merged this will fix #964

### DIFF
--- a/python/cib/gen_str_catalog.py
+++ b/python/cib/gen_str_catalog.py
@@ -441,7 +441,13 @@ def extract_enums(filename: str):
     for node in translation_unit.cursor.walk_preorder():
         if node.kind == CursorKind.ENUM_DECL:
             new_decl = {
-                e.spelling: e.enum_value
+                # workaround for https://github.com/llvm/llvm-project/issues/221326
+                e.spelling: 1
+                if (
+                    e.type.get_declaration().enum_type.spelling == "bool"
+                    and e.enum_value == -1
+                )
+                else e.enum_value
                 for e in node.walk_preorder()
                 if e.kind == CursorKind.ENUM_CONSTANT_DECL
             }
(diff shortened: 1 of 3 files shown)


elbeno (approved):

elbeno:
This is a better fix than my original weird workaround, thanks @jefftrull for tracking this down

elbeno:
@jefftrull The quality is failing because of python formatting. It can be fixed by:

```bash
$ uv venv
$ source .venv/bin/activate
$ uv sync
$ ninja -C build fix-ruff-format
```

(or at the last line, however you would like to build the `fix-ruff-format` target)

jefftrull:
Force-pushed a formatting fix but I'm not sure what to do about the `uv` issue (the one from the previous CI run, seemed like it wasn't found or something)

elbeno:
It was likely a transient issue. We'll see...

jefftrull:
Much appreciated!

PR #970, opened 2026-09-08
Normalize signed zero lookup keys

lukevalenty:
Normalize floating-point zero before hashing so +0 and -0 match.

### DIFF
--- a/include/lookup/pseudo_pext_lookup.hpp
+++ b/include/lookup/pseudo_pext_lookup.hpp
@@ -26,6 +26,7 @@ constexpr auto as_raw_integral(auto v) {
     static_assert(sizeof(v) <= 8);
 
     if constexpr (std::is_floating_point_v<decltype(v)>) {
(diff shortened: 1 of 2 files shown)


include/lookup/pseudo_pext_lookup.hpp
```diff
@@ -17,13 +17,20 @@
...
 
 namespace lookup {
 
 namespace detail {
 constexpr auto as_raw_integral(auto v) {
     static_assert(sizeof(v) <= 8);
 
+    if constexpr (std::is_floating_point_v<decltype(v)>) {
+        if (v == 0) {
+            v = 0;
```
bdeane-intel, line 31: **suggestion**: This is worth a comment.

bdeane-intel (approved):

include/lookup/pseudo_pext_lookup.hpp
lukevalenty: Added a comment explaining the normalization.

PR #975, opened 2026-09-16
:white_check_mark: Add tests for field matchers

elbeno:
Problem:
- There are no tests that actually check that the field matchers work when used by a field.
- https://github.com/intel/compile-time-init-build/pull/974

Solution:
- Add tests so that the behaviour is maintained.


lukevalenty:
😳

lukevalenty (approved):

PR #976, opened 2026-09-16
:art: Use consistent naming approach for message lib

elbeno:
Problem:
- The message library uses ad-hoc naming.

Solution:
- Use stdx naming approach.

Note:
- This is one of several PRs that will unify naming approaches in CIB.


lukevalenty (approved):
🙌🏻

